PR #14, opened 2024-09-23
:sparkles: Move and adjust properties from MTA extension manifest

rszwajko:


ibolton336 (approved):

vscode/resources/mta_mtr_icon.png
djzager, line 1: Since we are under the konveyor org, cncf owned, I believe this should be the konveyor logo. Maybe we use this one? https://github.com/konveyor-ecosystem/kai-vscode-plugin/blob/main/resources/konveyor-icon-color.png

djzager (changes requested):

vscode/resources/mta_mtr_icon.png
rszwajko: @djzager 
My understanding was that this plugin will replace the MTA plugin in the VS Code marketplace so we want to maintain some sense of continuity.

djzager: Then we would need to fork this repo for producing such a replacement plugin because we cannot put red hat product stuff in cncf owned repositories.

djzager: That is to say. I agree we will need to figure out how to make this plugin replace the MTA plugin, we just can't have product bits in these konveyor/* repositories.

dymurray: +1 on branding this under Konveyor right now. I fully support the idea to template out the rendering to make branding changes easy down the line

djzager (approved):

PR #15, opened 2024-09-24
Move repo level config from tackle2-ui

rszwajko:
Changes:
1. GH actions:
    -  PR template
    -  on PR/push action with 2 jobs: standard prod build and extension deployment using virtual screen
    -  nightly using the same jobs as PR/push action
2.  Add CI badges to main README
3. Add linting - rules inherited from VS Code generator and rules from tackle2-ui
4. Add OWNERS file
5. Fix test target - use separate build folder for webpack (out) and for tsc(test_out)
6. Set minimal node and npm versions - enforce them in GH actions via .nvmrc file
7. disable watch flag during webpack builds in prod mode
8. Fix linting errors detected by newly added linters

Reference-Url: https://github.com/konveyor/tackle2-ui/tree/release-0.5


.github/workflows/nightly-ci-repo.yaml
```diff
@@ -0,0 +1,11 @@
+name: Nightly CI (repo level @main)
+
+on:
+  schedule:
+    - cron: "35 5 * * *" # every day @ 5:35am UTC
```
djzager, line 5: Can we just add this to `ci-repo.yml`?

.github/workflows/ci-repo.yml
```diff
@@ -0,0 +1,82 @@
...
+          node-version-file: ".nvmrc"
+
+      - name: Cache extension node_modules
+        uses: actions/cache@v3
+        with:
+          path: vscode/node_modules
+          key: ${{ runner.os }}-node-${{ hashFiles('vscode/package-lock.json') }}
+
+      - name: Verify package-lock.json
+        run: ./scripts/verify_lock.mjs
```
djzager, line 42: Speaking from ignorance, do we need this script? Could we not just run `npm ci` and verify package-lock.json remains unchanged?

vscode/.gitignore
djzager, line 1: Nice. I hadn't realized this was a thing we could do (add `.gitignore` to subdir).

djzager (commented):
Looks good. Some comments for you.

.github/workflows/nightly-ci-repo.yaml
sjd78: Splitting it up helped with keeping the CI badges accurate on tackle2-ui.

.github/workflows/ci-repo.yml
sjd78: Short answer is cachito.

This script was created to make sure the lock file will still work when built downstream.  There is a weird edge case where the lockfile can be ok for npm, but not be ok for cachito.

If no downstream build .... it isn't really super useful, but is a nice to have.

vscode/.prettierrc.mjs
```diff
@@ -0,0 +1,15 @@
...
+  // default value used
+  // trailingComma: "all",
+  // semi: true,
+  // singleQuote: false,
+
+  // Values used from .editorconfig:
+  //   - printWidth == max_line_length
+  //   - tabWidth == indent_size
+  //   - useTabs == indent_style
+  //   - endOfLine == end_of_line
```
sjd78, line 12: Also need `.editorconfig`.  Copying from tackle2-ui would work.

Except ..... let's please have a wider `max_line_length`!  100 would be much nicer than 80.

vscode/eslint.config.mjs
```diff
@@ -1,28 +1,91 @@
...
+        { argsIgnorePattern: "^_" },
+      ],
+      "unused-imports/no-unused-imports": ["warn"],
+      "@typescript-eslint/no-explicit-any": "warn",
+      "react/jsx-key": "warn",
+      // "react-hooks/rules-of-hooks": "warn",
+      // "react-hooks/exhaustive-deps": "warn",
+      "no-extra-boolean-cast": "warn",
+      "prefer-const": "warn",
+      "react/no-unknown-property": ["warn"],
```
sjd78, line 76: Should adjust most of these back to error since they're warn more because the state of tackle2-ui's code.  Probably not as many issues in the current code base here. ;-)

rszwajko: VS Code API is loosely typed and rules like `no-explicite-any` would cause more harm then benefits if set to error level - check i.e. [Command interface](https://github.com/microsoft/vscode/blob/0f9abf3c3567e53df749299dafcc988831a3b4e7/src/vscode-dts/vscode.d.ts#L40)

.github/workflows/nightly-ci-repo.yaml
rszwajko: I've moved the badges to main README

.github/workflows/ci-repo.yml
rszwajko: @sjd78 @djzager I've run into problems with the script - mainly due to the fact we have kind-of-monorepo here. 
Since we have no downstream I've switched to `npm ci` for now.

.github/workflows/ci-repo.yml
```diff
@@ -0,0 +1,81 @@
...
+        uses: actions/cache@v3
+        with:
+          path: vscode/node_modules
+          key: ${{ runner.os }}-node-${{ hashFiles('vscode/package-lock.json') }}
+
+      - name: Install dependencies
+        run: |
+          cd vscode
+          npm version
+          npm ci
```
djzager, line 45: I imagine it might look a little less silly, although not practically different if you did this.

```suggestion
      - name: Install dependencies
        working-directory: ./vscode
        run: |
          npm version
          npm ci
```

djzager:
This looks good although still in draft so I'll save the approval until it's ready.

ibolton336 (approved):

.github/workflows/ci-repo.yml
rszwajko: That's also an option. Simple `cd vscode` was the fastest fix.  But there is more then one place where we have this kind of problems. We should probably decide how we want to proceed with our kind-of-monorepo.

rszwajko:
Example of GH runs on my fork - in particular a failed vscode test - https://github.com/rszwajko/editor-extensions/actions/runs/11020468337/job/30605307345?pr=2

djzager (approved):

PR #29, opened 2024-09-26
:sparkles: Add extension state and analysis placeholders

ibolton336:
- Add placeholder menu items for analyzing specific files from the explorer view 
- Add placeholder startAnalysis command 
- Create utility function for getting the webview content since it is being used in panel and sidebar webviews
- Create extension state to hold any new webview providers. Allows access to the state without having to change function signatures to pass in any new webviews to the registerCommand function as they are created.


djzager (approved):
I like it a lot. Nice.

Issue #11, opened 2024-09-23
vscode extension release automation


djzager:
We need to set up an automated process to package our VSCode extension and prepare it for publishing to both the VSCode Marketplace and the Open VSX Registry. This will ensure that every push to `main` and every new release results in a packaged extension, ready for installation and distribution.

### Tasks:

1. **Automated Packaging:**
   - Use **GitHub Actions** to automate packaging the extension on every push to the `main` branch and on every release.
   - We can leverage the `vsce` tool to handle the packaging of the extension.
     - [VSCE Documentation](https://code.visualstudio.com/api/working-with-extensions/publishing-extension)
     - [VSCE GitHub](https://github.com/microsoft/vscode-vsce)
   - GitHub Actions documentation for reference:
     - [Creating GitHub Actions workflows](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions)
     - [GitHub Actions for VSCode extensions](https://github.com/HaaLeo/publish-vscode-extension)

   **Steps to automate packaging:**
   - Set up a GitHub Actions workflow that runs on push to `main` and release events.
   - The workflow should:
     - Install dependencies (e.g., `npm install`).
     - Run `vsce package` to bundle the extension.
     - If required, create a release artifact for downloading the packaged `.vsix` file.

   Example GitHub Actions workflow:
   ```yaml
   name: Package VSCode Extension

   on:
     push:
       branches:
         - main
     release:
       types: [created]

   jobs:
     build:
       runs-on: ubuntu-latest

       steps:
         - name: Checkout code
           uses: actions/checkout@v2

         - name: Set up Node.js
           uses: actions/setup-node@v3
           with:
             node-version: '16'

         - name: Install dependencies
           run: npm install

         - name: Package extension
           run: npx vsce package

         - name: Upload VSIX as artifact (optional)
           uses: actions/upload-artifact@v2
           with:
             name: my-extension
             path: '*.vsix'
   ```

2. **Publishing to the VSCode Marketplace:**
   - Create issues to handle the steps required for publishing to the VSCode Marketplace.
   - Publishing requires a publisher account and a Personal Access Token (PAT) from Azure DevOps to authenticate with the marketplace.
   - Documentation for publishing to the VSCode Marketplace:
     - [VSCode Publishing Documentation](https://code.visualstudio.com/api/working-with-extensions/publishing-extension)
[...]

PR #34, opened 2024-09-30
:sparkles: Integrating release workflow - initial build, test, package, and upload artifacts to the ci runs.

savitharaghunathan:
This workflow builds the package, runs tests, and packages the extension. It also uploads it to the GH ci runs. It runs on merge and can be manually triggered on branches main and release-*. Supports linux, macos, windows.
ref issue: https://github.com/konveyor/editor-extensions/issues/11
todo:

- [ ] Add Konveyor-analyzer binaries
- [ ] Auto Publish it to the marketplace

Sample run - https://github.com/savitharaghunathan/editor-extensions/actions/runs/11111525940

I had to update the Readme as the `vsce package` fails if we use template generated readme file.


savitharaghunathan:
Ref issue: https://github.com/konveyor/editor-extensions/issues/11

djzager (changes requested):
This looks great. Thank you for making quick work of this. Only minor changes requested.

If you could please create an issue with all that you know to be required to publish to the VSCode marketplace and/or the Open VSX Registry, that would be fantastic.

.github/workflows/release.yml
```diff
@@ -0,0 +1,114 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - "main"
+      - "release-*"
+
+  pull_request:
```
shawn-hurley, line 9: I don't know if we want to release on a PR, unless I am missing something which very well might be the case

.github/workflows/release.yml
```diff
@@ -0,0 +1,114 @@
...
+      - "release-*"
+
+  workflow_dispatch:
+    inputs:
+      release_branch:
+        description: 'Branch to release (e.g., main, release-1.0)'
+        default: 'main'
+      run_tests:
+        description: 'Run tests before packaging and releasing (true/false)'
+        default: 'true'
```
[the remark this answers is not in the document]
savitharaghunathan: Ack. I can get rid of this.

.github/workflows/release.yml
savitharaghunathan: We were discussing to have a build for every pr, so that if anyone wants to test the extension can jut download and install.

savitharaghunathan: just*

savitharaghunathan: May be I need two workflows. One for PRs and one one for releases. Thanks for the question

.github/workflows/release.yml
```diff
@@ -0,0 +1,114 @@
...
+        working-directory: ./vscode
+        run: |
+          npm install @vscode/vsce
+          npx vsce package
+
+      - name: Upload VSIX artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: vscode-extension-${{ matrix.arch }}-${{ github.run_id }}
+          path: "vscode/*.vsix"
```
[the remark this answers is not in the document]
savitharaghunathan: I like that too. I will update the workflow. https://github.com/marvinpinto/action-automatic-releases has been archived. I will look into alternatives.

.github/workflows/release.yml
savitharaghunathan: also this only uploads the vsix package as artifacts to the ci runs. eg: https://github.com/savitharaghunathan/editor-extensions/actions/runs/11111525940. @shawn-hurley wdyt? we can chat more if you think it shouldnt be this way. May be I am missing something

shawn-hurley: I think that all makes a ton of sense! I very much think then that is a good idea and something that I would for sure use

.github/workflows/release.yml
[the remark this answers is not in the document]
savitharaghunathan: removed

.github/workflows/release.yml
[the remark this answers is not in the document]
savitharaghunathan: I tried two other alternatives but it is not updating the artifacts/commit. The alternative will be to create our scripts to remove the artifacts and then upload (marvinpinto's script: https://github.com/marvinpinto/actions/blob/master/scripts/mirror-actions.sh) 
@djzager which path do you suggest I pursue? use the archived action or write our own. I can write our own since I am not a fan of using unmaintained code.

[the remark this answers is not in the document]
djzager: We use https://github.com/ncipollo/release-action for the Konveyor release workflows...still looks like a legit process. It seems that this is valid and should work to maintain a single/rolling `latest` release (yes it is AI generated) :-D

```
      - name: Create or update a GitHub release
        uses: ncipollo/release-action@v1
        with:
          tag: latest
          name: Latest Release
          draft: false
          prerelease: false
          body: "Rolling release of the latest build from the main branch."
          files: |
            path-to-your-artifact-1
            path-to-your-artifact-2
          token: ${{ secrets.GITHUB_TOKEN }}
```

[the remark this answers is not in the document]
savitharaghunathan: I tried this one. It only updates the .vsix file and not the source code zip files. I also tried another one `softprops/action-gh-release@v1` which did the same thing.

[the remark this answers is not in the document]
savitharaghunathan: I will try again probably tomorrow. I might have missed something for sure.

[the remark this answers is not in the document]
djzager: Did you make use of `allowUpdates`?
> An optional flag which indicates if we should update a release if it already exists. Defaults to false.

I notice also:

removeArtifacts | Indicates if existing release artifacts should be removed. | false | false
-- | -- | -- | --
replacesArtifacts | Indicates if existing release artifacts should be replaced. | false | true

You may well have tried these already but we can talk about it more tomorrow. We may have to delete the release and recreate it if it is going to be difficult.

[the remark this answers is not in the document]
savitharaghunathan: I used both `allowUpdates` and `replacesArtifacts`. It only replaces the `.vsix` package and not the source code

.github/workflows/release.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Build
+        working-directory: ./vscode
+        run: npm run package
+
+      # Install Xvfb and Run tests on Linux
+      - name: Install Xvfb and Run tests (Linux)
+        if: matrix.arch == 'linux'
+        working-directory: ./vscode
+        run: |
+          sudo apt-get install -y xvfb
```
rszwajko, line 57: it seems `xvfb` is installed on the `ubuntu-latest` - at least it works of us in `ci-repo` workflow.

rszwajko (commented):
Looks good! One minor comment (see below) and one question: how long does the build take?  If it's quick then we should just replace the current `ci-repo` workflow (which is running only Linux and finishes on package step).  If it takes long (windows and macos!) then I would run it in `nightly-ci-repo` workflow (once a day).

.github/workflows/release.yml
savitharaghunathan: when I tried without that, the npm test failed. IIRC, it was related to some GUI component. I will take a look into the `ci-repo` to see if I am doing something wrong

savitharaghunathan:
> Looks good! One minor comment (see below) and one question: how long does the build take? If it's quick then we should just replace the current ci-repo workflow (which is running only Linux and finishes on package step). If it takes long (windows and macos!) then I would run it in nightly-ci-repo workflow (once a day).

@rszwajko it takes about ~2 mins to run all 3 builds. Sometimes the delay is because it is waiting for the runners.

rszwajko:
@savitharaghunathan 
> it takes about ~2 mins to run all 3 builds. Sometimes the delay is because it is waiting for the runners.
>
Then I would go with extending our ci-repo script (we will re-use those cool status badges then :) )

savitharaghunathan:
> @savitharaghunathan
> 
> > it takes about ~2 mins to run all 3 builds. Sometimes the delay is because it is waiting for the runners.
> 
> Then I would go with extending our ci-repo script (we will re-use those cool status badges then :) )

@rszwajko I can do that. We can always separate it out later if things get complex.

.github/workflows/release.yml
savitharaghunathan: updated it. Thanks for the catch

.github/workflows/release.yml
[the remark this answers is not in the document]
savitharaghunathan: @djzager I have updated the code to cut a prerelease for every merge using a name that is `dev-<commit-id_short>`. We can add a cleanup script or if this doesnt workout then I can work on the other approach of cleaning up latest before release. Let me know what you think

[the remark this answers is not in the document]
savitharaghunathan: eg: https://github.com/savitharaghunathan/editor-extensions/releases/tag/dev-4c084ab5

.github/workflows/ci-repo.yml
```diff
@@ -11,71 +11,98 @@ on:
...
+        with:
+          ref: ${{ github.ref }}
 
-      - name: Use Node.js from .nvmrc
+      - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
           node-version-file: "vscode/.nvmrc"
 
-      - name: Cache extension node_modules
```
rszwajko, line 35: I would keep the cache. The pattern of using a separate action is taken from [Continue workflows](https://github.com/continuedev/continue/blob/70d9bc5d52f328d94867a7eb08a41ace3ad06c17/.github/workflows/ts-check.yaml#L24). If you prefer we could also use caching via `setup-node` action.

savitharaghunathan: added it back. Thanks

.github/workflows/ci-repo.yml
```diff
@@ -11,71 +11,104 @@ on:
...
+        if: matrix.arch == 'linux' && github.ref == 'refs/heads/main'
+        uses: ncipollo/release-action@v1
+        with:
+          tag: ${{ steps.vars.outputs.tag_name }}
+          allowUpdates: true
+          commit: ${{ github.sha }}
+          artifacts: "vscode/*.vsix"
+          prerelease: true
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
djzager, line 114: I think this should be it's own build that:
1. Runs only for `refs/heads/main`
2. `needs: package`
2. Assuming that `tag` hasn't been the source of our problem, then we should make this a semantic version compatible string like `v0.0.1-dev+${{ github.sha }}` (you are welcome to shorten it if it looks terrible).

djzager (changes requested):
I really like moving this into ci-repo. A few comments.

.github/workflows/ci-repo.yml
```diff
@@ -11,71 +11,104 @@ on:
       - "main"
       - "release-*"
 
-  workflow_dispatch:
-
-  workflow_call:
```
djzager, line 16: I suspect we'll need to keep this in order for the nightlies to work.

savitharaghunathan: added it back

.github/workflows/ci-repo.yml
savitharaghunathan: @dzager: do you know where can I get the release version from? do I need to add an env variable or is there any other way to know?

.github/workflows/ci-repo.yml
```diff
@@ -11,71 +11,104 @@ on:
...
+          npx vsce package
 
-      - name: Use Node.js from .nvmrc
-        uses: actions/setup-node@v4
+      - name: Upload VSIX artifact
+        uses: actions/upload-artifact@v4
         with:
-          node-version-file: "vscode/.nvmrc"
+          name: vscode-extension-${{ matrix.arch }}-${{ github.run_id }}
+          path: "vscode/*.vsix"
```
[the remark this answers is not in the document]
savitharaghunathan: @djzager I have separated the jobs out. Let me know if it makes sense. It takes about ~5min to run now

.github/workflows/ci-repo.yml
```diff
@@ -10,24 +10,36 @@ on:
     branches:
       - "main"
       - "release-*"
-
   workflow_dispatch:
 
   workflow_call:
```
djzager, line 15: Please update these so we can use them to dictate when we publish.
```suggestion
  workflow_dispatch:
    inputs:
      shouldRelease:
        description: 'Should a release be created?'
        required: false
        default: 'false'

  workflow_call:
    inputs:
      shouldRelease:
        description: 'Should a release be created?'
        required: false
        default: 'false'
```

.github/workflows/ci-repo.yml
```diff
@@ -39,43 +51,135 @@ jobs:
...
-          path: vscode/node_modules
-          key: ${{ runner.os }}-node-${{ hashFiles('vscode/package-lock.json') }}
+          name: vscode-extension-${{ matrix.arch }}
+          path: "vscode/*.vsix"
+
+  release:
+    name: Release
+    runs-on: ubuntu-latest
+    needs: package
+    if: github.ref == 'refs/heads/main'
```
djzager, line 157: Update this to run on shouldRelease
```suggestion
    if: |
      github.ref == 'refs/heads/main' &&
      (github.event.inputs.shouldRelease == 'true' || github.event.inputs.shouldRelease == true)
```

.github/workflows/ci-repo.yml
```diff
@@ -39,43 +51,135 @@ jobs:
...
+        with:
+          path: ./artifacts
+
+      - name: Generate tag name
+        id: vars
+        shell: bash
+        run: |
+          SHORT_SHA=$(echo "${GITHUB_SHA}" | cut -c1-8)
+          TAG_NAME="v0.0.1-dev-${SHORT_SHA}"
+          echo "tag_name=$TAG_NAME" >> $GITHUB_OUTPUT
```
djzager, line 174: We can do this so we actually generate a legit semver release that moves about.

```suggestion
      - name: Generate tag name with timestamp
        id: vars
        shell: bash
        run: |
          TAG_NAME="v0.0.1-dev+$(date +%Y%m%d)"
          echo "tag_name=$TAG_NAME" >> $GITHUB_OUTPUT
```

.github/workflows/ci-repo.yml
```diff
@@ -39,43 +51,135 @@ jobs:
...
+
+      - name: Create Pre-release
+        uses: ncipollo/release-action@v1
+        with:
+          tag: ${{ steps.vars.outputs.tag_name }}
+          releaseName: "Development Build - ${{ steps.vars.outputs.tag_name }}"
```
djzager, line 180: Now that we are just going to run this on nightlies, and they'll be unique, we can drop the releaseName (and just let it take the tag name).

djzager (changes requested):
I think this is the last set of changes I'll need to ask for. Thank you for your patience, I really had not expected that we could make this so consumable for other stakeholders this quickly...really great work.

My requested changes reflect what I think to be the best option. You should also update the nightly workflow like:

```
jobs:
  nightly:
    uses: ./.github/workflows/ci-repo.yml
    with:
      shouldRelease: true
```

That will give us `v0.0.1-dev+20241003` releases as early as tomorrow night that can with downloadable vscode extensions our stakeholders can install and use. Exceeds all of my expectations.

.github/workflows/ci-repo.yml
djzager: If you think it should be `nightly` instead of `dev`, that's cool, I just don't think we should make it variable at this stage.

.github/workflows/ci-repo.yml
```diff
@@ -154,23 +166,33 @@ jobs:
     name: Release
     runs-on: ubuntu-latest
     needs: package
-    if: github.ref == 'refs/heads/main'
+    if:
```
djzager, line 169: May need this:
```suggestion
    if: |
```

.github/workflows/ci-repo.yml
```diff
@@ -154,23 +166,33 @@ jobs:
...
 
       - name: Download VSIX artifacts
         uses: actions/download-artifact@v3
         with:
           path: ./artifacts
 
-      - name: Generate tag name
+      - name: Read version from package.json
+        id: get_version
+        run: echo "version=$(node -p \"require('./package.json').version\")" >> $GITHUB_OUTPUT
```
djzager, line 189: Do you need to set the working directory?

djzager (dismissed):
Nice! A few formatting comments that may be impactful.

.github/workflows/ci-repo.yml
savitharaghunathan: good catch. I definitely need to since we are using the pattern from continue

.github/workflows/ci-repo.yml
```diff
@@ -11,71 +11,104 @@ on:
...
+        if: matrix.arch == 'macos'
+        working-directory: ./vscode
+        run: npm test
+
+      # Run tests on Windows
+      - name: Run tests (Windows)
+        if: matrix.arch == 'windows'
+        working-directory: ./vscode
+        shell: cmd
+        run: npm test
```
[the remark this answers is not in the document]
savitharaghunathan: updated

.github/workflows/ci-repo.yml
savitharaghunathan: @djzager I fetched the version from the package.json but I am not able to test it on my fork ( messed up main branch :( )

.github/workflows/ci-repo.yml
savitharaghunathan: updated

.github/workflows/ci-repo.yml
savitharaghunathan: I am cool with dev.

.github/workflows/ci-repo.yml
savitharaghunathan: Thanks for catching this

.github/workflows/ci-repo.yml
savitharaghunathan: removed releaseName

.github/workflows/ci-repo.yml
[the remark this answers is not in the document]
savitharaghunathan: updated

savitharaghunathan:
@djzager ptal when you get some time. I am not able to test the release step on my fork. do you have any thoughts? Also, what do you mean by this -

> Nice! A few formatting comments that may be impactful.

comments for steps?

djzager (approved):
LGTM

PR #36, opened 2024-10-01
:ghost: Remove OWNERS file

rszwajko:


djzager:
The question is if, adding contributors with write permissions and preventing anyone with <write permissions from approving PRs is sufficient to prevent us from having to have a OWNERS.md and CODEOWNERS.md.

CODEOWNERS
```diff
@@ -0,0 +1,4 @@
+#
```
djzager: Since our branch protection rules works on the configured contributors with write access, this file is not required.

djzager (approved):

PR #40, opened 2024-10-02
:sparkles: Adds mock analysis results & populate webview with patternfly react datalist

ibolton336:
- Generate mock data for incidents
- Add diagnostics
- Add incident data to workspace state
- Reset / sync state between diagnostics and webview
- Add in patternfly
- Update views to use patternfly


vscode/src/commands.ts
```diff
@@ -11,36 +14,79 @@ function getFullScreenTab() {
...
         const contentString = Buffer.from(fileContent).toString("utf8");
 
-        console.log(contentString, fileContent);
-
-        // TODO: Analyze the file content
-        vscode.window.showInformationMessage(`Analyzing file: ${filePath}`);
+        await extensionContext.workspaceState.update(
+          "incidentData",
+          JSON.stringify(generateMockIncidentData()),
+        );
```
djzager, line 87: My original thought was that whatever process performs analysis would pass `[]Ruleset` ([found here](https://github.com/konveyor/analyzer-lsp/blob/8f5ccf72040d7d89765aa3b8fc86e8d7cb4be684/output/v1/konveyor/violations.go#L16))as a parameter, and do the parsing there.

I'm wondering if you are thinking it would work differently...maybe the analysis process would pass the `[]RuleSet` to a "process function" that put it in the workspace state where it got added to diagnostics (etc. etc.).

vscode/src/webviewMessageHandler.ts
```diff
@@ -1,16 +1,43 @@
...
+  provider: KonveyorGUIWebviewViewProvider,
+) {
+  const incidentDataString = state.extensionContext.workspaceState.get<string>(
+    "incidentData",
+    "[]",
+  );
+  const incidentData = JSON.parse(incidentDataString);
+
+  webview.onDidReceiveMessage(async (message) => {
+    const fileUri = vscode.Uri.parse(message.file);
```
djzager, line 17: This is really cool to see. Should this particular line be in the `openFile` case to prevent failure when the `message.file` would not be defined?

vscode/src/webview/types.ts
djzager, line 1: I figured we might mirror the analyzer-lsp types here...something like (chatgpt generated):

```
interface RuleSet {
    name?: string;
    description?: string;
    tags?: string[];
    violations?: Record<string, Violation>;
    insights?: Record<string, Violation>;
    errors?: Record<string, string>;
    unmatched?: string[];
    skipped?: string[];
}

interface Violation {
    description: string;
    category?: Category;
    labels?: string[];
    incidents: Incident[];
    links?: Link[];
    extras?: any;
    effort?: number;
}

type Category = "potential" | "optional" | "mandatory";

interface Incident {
    uri: string;
    message: string;
    codeSnip?: string;
    lineNumber?: number;
    variables?: Record<string, any>;
}

interface Link {
    url: string;
    title?: string;
}
```

vscode/src/webview/components/IncidentList.tsx
djzager, line 1: 🚀 Nice to see this.

vscode/src/webview/components/App.tsx
```diff
@@ -1,35 +1,69 @@
...
+        ) : (
+          <EmptyState>
+            <EmptyStateIcon icon={SearchIcon} />
+            <Title headingLevel="h4" size="lg">
+              No Incidents Found
+            </Title>
+            <EmptyStateBody>Run an analysis to see results here.</EmptyStateBody>
+            <Button
+              variant="primary"
+              onClick={() => vscode.postMessage({ command: "startAnalysis" })}
```
djzager, line 59: What's happening here. I now notice it all over the place and in the cat coding webview example. Is this actually running the command or doing something else?

djzager (commented):
This is awesome to see this coming together. Some hopefully helpful comments. Need more time to review the mock data and how we are using the state, looks interesting.

vscode/src/commands.ts
ibolton336: This definitely help me understand what we are looking for. Need to think on it a bit more but I can make the change and have the ruleset be passed in.

vscode/src/webview/components/App.tsx
ibolton336: Ah it is just passing back a block of data from webview to extension listener... the command term is probably overloaded here.

vscode/src/webview/mockData.ts
djzager, line 1: Should we be keeping this around?

vscode/src/webview/types.ts
djzager, line 1: Nice!

djzager (approved):

vscode/src/webview/types.ts
```diff
@@ -0,0 +1,86 @@
...
+  labels?: string[];
+
+  // Incidents list of instances of violation found
+  incidents: Incident[];
+
+  // ExternalLinks hyperlinks to external sources of docs, fixes
+  links?: Link[];
+
+  // Extras reserved for additional data
+  extras?: any; // Using 'any' as a replacement for json.RawMessage
```
rszwajko, line 60: I would go with `unknown` and use type guards to narrow the type where needed.
Alternatively we could go with an equivalent of [`LSPAny`](https://github.com/microsoft/vscode-languageserver-node/blob/8c5b095be0bb0fa28dac28f4d97512e489b38884/types/src/main.ts#L90) which is used i.e. for [`diagnostic.data`](https://github.com/microsoft/vscode-languageserver-node/blob/8c5b095be0bb0fa28dac28f4d97512e489b38884/types/src/main.ts#L745)

vscode/src/webview/types.ts
```diff
@@ -0,0 +1,86 @@
+export interface Incident {
```
rszwajko, line 1: It seems "our" version of the [`Diagnostic`](https://github.com/microsoft/vscode-languageserver-node/blob/8c5b095be0bb0fa28dac28f4d97512e489b38884/types/src/main.ts#L688).
Do we want to align them more?

vscode/src/commands.ts
```diff
@@ -11,48 +13,59 @@ function getFullScreenTab() {
   return tabs.find((tab) => (tab.input as any)?.viewType?.endsWith("konveyor.konveyorGUIView"));
 }
 
-const commandsMap: (
-  extensionContext: vscode.ExtensionContext,
-  state: ExtensionState,
-) => {
+const commandsMap: (state: ExtensionState) => {
```
rszwajko, line 16: if `any` is required here (i.e. due to vs code interface) then let's silence the linters  - GH actions create a lot of noise here (the msg is multiplied by platforms):
image

rszwajko (commented):
@ibolton336  please fix the warnings (resolve or silence) because the GH generated warnings make the code hard to review (warning are multiplied by platforms - see below). The rest seems OK to me - we can merge to unblock others and refine it later on.

vscode/src/webview/types.ts
ibolton336: I'd like to move toward this - yes. Would like to iterate further on this but currently the Incident type covers what we are receiving from Kantra.

djzager (dismissed):

djzager (dismissed):

rszwajko (approved):

PR #51, opened 2024-10-11
:seedling: Prerelease prune

savitharaghunathan:
This workflow deletes prerelease and associated tags older than 7 days. 

@djzager @dymurray do you think we need the prerelease artifacts more than a week?

### DIFF
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -3,7 +3,8 @@ name: Prune Nightly Pre-releases
 on:
   schedule:
     - cron: "0 6 * * *"  # Runs every day at 6:00am UTC
-
+  workflow_dispatch:
+  
 jobs:
   prune:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


djzager:
I haven't really given a lot of thought to this. Having a rolling window doesn't seem to be a bad idea but I am also somewhat afraid to throw away the older bits.

What do you think about running the workflow once per week, say Sunday, throwing away all but the newest?

savitharaghunathan:
> I haven't really given a lot of thought to this. Having a rolling window doesn't seem to be a bad idea but I am also somewhat afraid to throw away the older bits.
> 
> What do you think about running the workflow once per week, say Sunday, throwing away all but the newest?

One of the reasons I kept a rolling window is to see if we need time to test a build or two. 
Are you saying we run it on every sunday and retain only the previous week's builds? that is sunday date - 7?

savitharaghunathan:
also with every day nightly release, whenever someone fetches upstream its gonna be a lot of tags if we dont purge. personally, I think it would be not helpful to see that many tags of prerelease version. May be I am wrong when I just fetch the whole of upstream and not everyone does that way.

PR #56, opened 2024-10-14
:sparkles: PF6 upgrade & new analyzer integration

ibolton336:


vscode/assets/bin/kai-analyzer
djzager, line 1: should drop this because we are using the `kai-analyzer.${os}.${arch}` builds.

vscode/src/VsCodeExtension.ts
```diff
@@ -48,6 +48,100 @@ export class VsCodeExtension {
       setupWebviewMessageListener(webview, this.state);
     });
 
+    // // DEBUG USE ONLY
```
djzager, line 51: Should just delete this blob right? Any reason to keep it?

vscode/src/webview/components/ChatbotContainer.tsx
djzager, line 1: neat.

djzager (changes requested):
A few things need to get cleaned up. Otherwise it looks good.

vscode/src/VsCodeExtension.ts
ibolton336: It's pretty useful for webview development. Maybe we can keep until we have time to implement something else like MSW for mock data.

vscode/src/webviewMessageHandler.ts
```diff
@@ -43,12 +43,7 @@ export function setupWebviewMessageListener(webview: vscode.Webview, state: Exte
 
         const folderUri = await vscode.window.showOpenDialog(options);
         if (folderUri && folderUri[0]) {
-          vscode.commands.executeCommand("konveyor.startAnalysis", folderUri[0]);
-        } else {
-          webview.postMessage({
-            type: "analysisFailed",
-            message: "No folder selected for analysis.",
-          });
+          vscode.commands.executeCommand("konveyor.runAnalysis", folderUri[0]);
```
djzager, line 46: I would skip the folder stuff and just call `runAnalysis`

djzager (approved):

djzager (approved):

PR #71, opened 2024-10-17
:sparkles: NPM workspaces and Migrate to Vite for HMR in Webview Package

### LINKED ISSUE #77: [vscode] dev mode flag so we have a reusable mechanism for faking/mocking out hard stuff
Thinking about our need to mock the request/response from kai/rpc-server because we haven't integrated it has become apparent that we may actually want to fake this most of the time so that we don't actually need to talk to the LLM unless we really want to.

We need a mechanism in our extension to put ourselves in a dev mode of sorts where we skip the long running calls so that we can iterate faster.

ibolton336:
PR changes: 
- Replace Webpack with Vite for webview-ui. In development mode, the extension is able to consume the webview running on a local vite dev server which allows for HMR and a rapid development experience. 
- Adds npm workspaces implementation for dependency sharing (typescript)  & better handling of build / dev scripts for development / production. Lifts all eslint & prettier configuration to project root. 
  - Creates a new `shared` package that holds shared code between webview-ui and extension. 
  - Can now run npm install && npm run dev from the project root to start both development servers (webpack for extension & vite for ui)
  - Vite enables HMR for webview development. Also removes need for complex plugin configuration - only plugin needed now is the [vite react plugin
](https://github.com/konveyor/editor-extensions/pull/71/files#diff-223561f5d3cf9a2528977c12fbd3b80c758ac2bf700fa6d85ff7071236178e30R2)
- Adds a new debugger pre launch task to start dev servers. This allows us to start all development servers & open the debugger with f5. 

Fixes #77 

TODO:
- Replace webpack copy plugin with something more robust
- Address vscode extension state management
- Improve webview messaging


sjd78 (commented):
Does adopting vite require setting up webview-ui as a separate project under vscode?  That is a bit concerning for knowing how to work in the project.

What does the future look like?  The webview as it's own project and then any regular extension points go in the base vscode tree?

Would setting up a more typical multi-workspace project make more sense?  Something like how tackle2-ui is setup?  The root project has build related configs, then common is stuff shared between webview and "normal" extensions, then webview, then "normal" extensions?

djzager (changes requested):
I don't see anything technically wrong with this. With it we need to document how this changes the development workflow. Either the vscode README.md or contributing doc or some combination of the two.

vscode/README.md
```diff
@@ -57,9 +57,27 @@ You can author your README using Visual Studio Code. Here are some useful editor
 - Toggle preview (`Shift+Cmd+V` on macOS or `Shift+Ctrl+V` on Windows and Linux).
 - Press `Ctrl+Space` (Windows, Linux, macOS) to see a list of Markdown snippets.
 
+## Developing & Running the extension
+
+```bash
```
ibolton336, line 62: @djzager

ibolton336:
> Does adopting vite require setting up webview-ui as a separate project under vscode? That is a bit concerning for knowing how to work in the project.
> 
> What does the future look like? The webview as it's own project and then any regular extension points go in the base vscode tree?
> 
> Would setting up a more typical multi-workspace project make more sense? Something like how tackle2-ui is setup? The root project has build related configs, then common is stuff shared between webview and "normal" extensions, then webview, then "normal" extensions?

The idea is here is based off the continue approach where they have a separate [GUI project](https://github.com/continuedev/continue/tree/main/gui) that remains separate from the extension code. Still working out what that looks like for us - will need to rethink CI with your input when the time is right.

vscode/tsconfig.json
```diff
@@ -1,14 +1,17 @@
 {
   "compilerOptions": {
-    "module": "Node16",
-    "target": "ES2023",
-    "lib": ["ES2023", "WebWorker", "DOM"],
-    "jsx": "react",
+    "module": "commonjs",
```
sjd78, line 3: Is that actually right?  Does vite output esm or commonjs for the project?

sjd78:
> The idea is here is based off the continue approach where they have a separate [GUI project](https://github.com/continuedev/continue/tree/main/gui) that remains separate from the extension code. Still working out what that looks like for us - will need to rethink CI with your input when the time is right.

I understand what you're trying to do now.  That is a great bit of info to put in the PR description

Should the webview/gui project then live outside of the vscode project tree?  That way it is obvious that the webview is intended to be independent of the vscode work.

vscode/tsconfig.json
sjd78: Actually I don't really even see where the not webview-ui stuff is being compiled/packaged...

ibolton336: Thats being done via the task.json using the tsc build. We still don't have a bundler with this approach but [esbuild](https://github.com/continuedev/continue/blob/main/extensions/vscode/scripts/esbuild.js) could work for us.

ibolton336: Ended up moving back to webpack for now & following your suggestion to moving webview to its own dir. Next step would definitely be to use a workspace or something similar to tackle2-ui.

djzager (approved):

vscode/src/extension.ts
```diff
@@ -1,45 +1,93 @@
...
-        }
-      });
+  constructor(context: vscode.ExtensionContext) {
+    this.state = {
+      analyzerClient: new AnalyzerClient(context),
+      sharedState: new SharedState(),
+      webviewProviders: new Set<KonveyorGUIWebviewViewProvider>(),
+      sidebarProvider: undefined as any,
+      extensionContext: context,
+    };
```
sjd78, line 18: I don't know much about how vscode extensions handle things -- is this the normal way to handle state?

I wonder if using https://immerjs.github.io/immer/ would make any kind of sense.

vscode/src/client/analyzerClient.ts
sjd78, line 1: In the `getSolution` function, prepending an underscore the argument names will clear up the warnings.

```ts
  public async getSolution(_webview: vscode.Webview, _incident: Incident): Promise<any> {
```

.gitignore
```diff
@@ -6,3 +6,6 @@ vscode/tsbuildinfo
 vscode/.metadata
 vscode/org.eclipse*
 .aider*
+node_modules
+webview-ui/node_modules
+webview-ui/build
```
sjd78, line 11: Should be enough to do just:
```
node_modules/
build/
```
and have directories with those names ignored by git from this folder down

sjd78 (commented):
This looks very close to being able to open the repo root directory in vscode and then work on both the `vscode/` extension and the `webview/` component all at the same time.

Running things is kinda interesting given the target.  So really I see a bunch of things going on:
  1. Hitting the run button in vscode to launch another vscode with the extension running (and the webview embedded in it)
  2. Doing `npm run` and having vscode with the extension running
  3. Doing a build to static files to be packaged
I'll need to tinker with what is here to see which one of those things work etc.  Working with running a command and then also hitting the run button is clunky but probably works ok for now.  Feels like it could be streamlined.

Beyond that, it would be good to see the messages that get sent between the webview and the extension be well defined somewhere like "webview-api" or something.  No better name comes to mind at the moment.  Then the webview would grab an instance of that object/whatever and the actions would be all well defined.  It is quite interesting to conceptualize the webview as an iframe in a browser that only has message posting access.  I'm not sure that's actually the case here.

I will look again in the morning with coffee.

ibolton336:
> * Hitting the run button in vscode to launch another vscode with the extension running (and the webview embedded in it)

This one was a head scratcher - couldn't get vscode debugger to open the launch.json unless I opened the nested vscode dir. Wanted to be able to launch the debugger from the project root, so had to create a new .vscode/launch.json in the project root.

.vscode/launch.json
```diff
@@ -0,0 +1,40 @@
...
+    {
+      "name": "Run Extension",
+      "type": "extensionHost",
+      "request": "launch",
+      "args": ["--extensionDevelopmentPath=${workspaceFolder}/vscode"],
+      "outFiles": ["${workspaceFolder}/vscode/out/**/*.js"],
+      "env": {
+        "VSCODE_DEBUG_MODE": "true"
+      },
+      "preLaunchTask": "npm: dev"
```
ibolton336, line 13: @sjd78 Just updated per our discussion: f5 is all we need apart from npm install to start the dev servers & open the  debugger window.

djzager (approved):
lgtm

djzager (approved):

PR #75, opened 2024-10-18
:seedling: introduce kai types needed for initialize and getSolutions

### LINKED ISSUE #66: [vscode] typescript types for the kai/rpc-server request/response for `getCodeplanAgentSolution`
We need to make the types for these so that we can mock them out.

Definition of Done:
- we have typescript types in "webview/types" or other appropriate place that represent the request we would send to the kai/rpc-server and the response we would get back.
- analyzerClient functions stubbed out for all of initialize, getCodeplanAgentSolution, shutdown, and exit

djzager:
Fixes #66

### DIFF
--- a/vscode/src/client/types.ts
+++ b/vscode/src/client/types.ts
@@ -1,0 +1,39 @@
+import { Incident } from "../webview/types";
+
+// KaiConfigModels type definition
+export interface KaiConfigModels {
+  provider: string;
+  args: Record<string, any>;
+  template?: string;
+  llamaHeader?: boolean;
+  llmRetries: number;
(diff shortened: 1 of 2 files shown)


vscode/src/client/types.ts
djzager, line 1: This is the stuff I would like for @fabianvf and @JonahSussman to double check to make sure I'm on the right track for what we should be sending/receiving from kai/rpc-server.

This will let us mock this out and iterate on the experience without having to actually talk to an LLM most of the time.

pranavgaikwad (approved):

ibolton336 (approved):

PR #76, opened 2024-10-18
:seedling: add start/stop/restart logic for server

djzager:
I don't know, maybe this is better pushed into the menubar or something, but I figured the plumbing for handling this in the view was the hard part.

### DIFF
--- a/vscode/src/commands.ts
+++ b/vscode/src/commands.ts
@@ -23,7 +23,32 @@ const commandsMap: (state: ExtensionState) => {
       }
 
       vscode.window.showInformationMessage("Starting analyzer...");
-      analyzerClient.start();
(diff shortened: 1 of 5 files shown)


vscode/src/commands.ts
```diff
@@ -23,7 +23,32 @@ const commandsMap: (state: ExtensionState) => {
       }
 
       vscode.window.showInformationMessage("Starting analyzer...");
-      analyzerClient.start();
+      if (fullScreenPanel && fullScreenPanel.webview) {
+        analyzerClient.start(fullScreenPanel.webview);
+      } else {
+        analyzerClient.start(state.sidebarProvider.webview!);
+      }
```
djzager, line 30: wonder if we should make a helper that'll return the webview to make this easier

Issue #72, opened 2024-10-17
[vscode] accept/reject a particular fix


djzager:
When a user accepts a fix, the change should be applied to the file and (probably) written to disk.

Definition of Done:
1. When user accepts a fix, it is applied to that file in the active workspace.

rszwajko:
Implemented in https://github.com/konveyor/editor-extensions/pull/93

Issue #70, opened 2024-10-17
[vscode/editor] in editor view of specific proposed changes


djzager:
When we are presenting the user with the response we got from the kai/rpc-server, one thing that would be really helpful is if the user could gain additional context on the proposed change by, clicking the filename for the codeblock or clicking the codeblock itself, opening that file in the editor where the proposed change is to be made. Additionally, we may elect to show the change in the editor with the option to accept/reject.

djzager:
Now that we have removed #69 from the minimum demoable thing, the scope of this effort has increased along with https://github.com/konveyor/editor-extensions/issues/67 , specifically this comment https://github.com/konveyor/editor-extensions/issues/67#issuecomment-2443145629.

The important notes about how this effort has changed:
1. The webview will message the diff viewer with the proposed solution after kai/rpc-server has responded to our solution request and we have performed any necessary post processing.
1. The webview will be "disabled", preventing user input while the user works through the proposed solution.
1. The diff-viewer must notify the webview when the user has "finished" reconciling the proposed solution.

rszwajko:
Most of the scope implemented in https://github.com/konveyor/editor-extensions/pull/79
The missing part is implementing the accept/reject buttons in the editor itself (i.e. code lense style). Right now it's implemented in the Konveyor Resolution view (native VS Code tree view).

rszwajko:
Missing parts implemented in #102

djzager:
I think this is done. Can we close this @rszwajko ?

Issue #68, opened 2024-10-17
[vscode] mock analyzerClient.getSolution


djzager:
Once we know the request/response we are going to send/receive to kai/rpc-server to `getCodeplanAgentSolution` in order to generate a solution for a particular violation, we should be able to mock or fake. This could be as simple as adding comments or removing lines, just something that would allow us to reliably exercise what the response would look like so we can develop the user experience for accept/rejecting fix.

rszwajko:
Completed with #91

PR #79, opened 2024-10-22
:sparkles: Create in-editor diff view

rszwajko:
Use built-in vscode.diff editor to compare the current code(pointed via
file URI) with proposed changes(pointed via custom konveyorMemFs Uri).

Key points:
1. proposed changes (diffs) are applied to current code base and
   are written to in-memory file system. Late they can be accessed
   via custom URI scheme konveyorMemFs.
2. both left and right side of the diff is editable
3. available commands grouped using `konveyor` (global) 
   or `konveyor.diffView` prefix(diff view specific):
   a) applyAll - apply all fixes
   b) revertAll - discard local changes to the proposed fixes
   c) applyFile/revertFile - as above but on a single file
   d) copyDiff/copyPath - convenience commands (single file scope)
   e) next/prev - keyboard navigation (for now) using the same key
      bindings as reference-view extension (f4 and shift+f4)
   f) viewFix - wrapper around vscode.diff that uses konveyorMemFs

Sources:
1. tree view is based on built-in reference-view extension. Only
   file-item nodes are used. Support for history items and dynamic model
   changes have been dropped. However the original architecture is there
   and should allow adding more advance features in the future.
2. in-memory file system is taken as-it-is from fsprovider-sample
3. the approach (using hidden file tree for fixes) originates from
   Continue's DiffManager
5. check/discard icons taken from source-control-sample

Reference-Url: https://github.com/continuedev/continue/blob/e7fe5994ffc6a3f45ad358ced7d6890deb145d96/extensions/vscode/src/diff/horizontal.ts#L34
Reference-Url: https://github.com/microsoft/vscode/tree/main/extensions/references-view
Reference-Url: https://github.com/microsoft/vscode-extension-samples/tree/main/fsprovider-sample
Reference-Url: https://github.com/microsoft/vscode-extension-samples/tree/main/source-control-sample


djzager (commented):
This is really cool to see what is possible. Thank you for getting this up so quickly.

I'm still trying to digest all that is going on here. My initial thought was that we would develop the mock response from the `kai/rpc-server::getCodeplanAgentSolution` and then look at how those change(s) get rendered, like diff viewer.

Things I don't yet understand:
* Do we need the tree view to navigate these things?
* How the result of our analyzerClient's `getSolution` for a specific incident is going to integrate with what is demonstrated here.

Things I'm uneasy about:
* Would prefer all of the commands to be registered in `commands.ts`. I really don't want them spread about the repository.
* The number of new commands being added. I would have thought we could have something like `konveyor.acceptFix` and `konveyor.rejectFix` and that is it.
* We are generating all of these file changes at startup...shouldn't this be tied to some user action? This is where I had thought the mocks would come into play...You call "Generate Fix" and you get changes that you can optional accept or reject.

rszwajko:
> * Do we need the tree view to navigate these things?
> 

No. Actually we use the tree as a list. However it's easier (and future-proof) to use tree view. The extensions I've followed ([reference-view](https://github.com/microsoft/vscode/blob/61ba31d9af8d633bfdbede3286c30922feb0bfc3/extensions/references-view/src/tree.ts#L13) and [SCM](https://github.com/microsoft/vscode/blob/61ba31d9af8d633bfdbede3286c30922feb0bfc3/src/vs/workbench/contrib/scm/browser/scmViewPane.ts#L2015)) use this approach. In SCM (i.e. git)  you can switch between tree and list but technically it's the same tree.

> * How the result of our analyzerClient's `getSolution` for a specific incident is going to integrate with what is demonstrated here.
> 

The solutions need to be written to [in-memory file system](https://github.com/rszwajko/editor-extensions/blob/e8985c599b7f005c81379e9b60f410866b1d6da2/vscode/src/diffView/register.ts#L10). Right now I'm using static mock data from `vscode/src/diffView/suggestions`. This approach originates from Continue's [DiffManager](https://github.com/continuedev/continue/blob/e7fe5994ffc6a3f45ad358ced7d6890deb145d96/extensions/vscode/src/diff/horizontal.ts#L34) but instead of real file system we are using in-memory one borrowed from [fsprovider-sample](https://github.com/microsoft/vscode-extension-samples/blob/main/fsprovider-sample/README.md)

>   * Would prefer all of the commands to be registered in `commands.ts`. I really don't want them spread about the repository.
> 

All commands are already listed in `package.json`.  We could duplicate that in `commands.ts` but this will require imports from all modules.

>   * The number of new commands being added. I would have thought we could have something like `konveyor.acceptFix` and `konveyor.rejectFix` and that is it.
> 

Most of the commands are related to navigation. `Copy*` were inherited from reference-view - the seem to be commonly used in VS Code in file views. 

>   * We are generating all of these file changes at startup...shouldn't this be tied to some user action? This is where I had thought the mocks would come into play...You call "Generate Fix" and you get changes that you can optional accept or reject.
>

Yep. Right now we are using static mock data. As soon as we will have better mocks we can switch.

rszwajko:
image

rszwajko:
@djzager
the current solution is not fully functional as it uses mock data however this will be improved in follow-up issues. First one is https://github.com/konveyor/editor-extensions/issues/68 which should provide us better mocks.
If you prefer to put some fixes already in this PR then please comment.

vscode/src/extension.ts
```diff
@@ -34,6 +35,7 @@ export function activate(context: vscode.ExtensionContext) {
         }
       });
   }
+  registerDiffView(context);
```
djzager, line 38: This illustrates what I believe to be my problem with this PR as currently constructed. It takes our extension down a completely different path with a whole new tree view and 8 new commands that are not readily tied into the work that @ibolton336 has done.

As an example, in this registerDiffView you define the command for `konveyor.diffView.applyAll`, so this PR would necessitate an entirely new command to be created for fixes to be accepted from the webview. This is not the direction I believe we should be heading in.

vscode/src/diffView/fileModel.ts
```diff
@@ -0,0 +1,222 @@
...
+  async asCopyText() {
+    let result = "";
+    for (const item of this.items) {
+      result += `${await item.asCopyText()}\n`;
+    }
+    return result;
+  }
+}
+
+export class KonveyorTreeDataProvider implements vscode.TreeDataProvider<FileItem> {
```
djzager, line 148: I reject the idea that we need a treedataprovider at this stage, the navigation of problems to solve and the files changed by a particular solution from kai/rpc-server should be in the webview.

vscode/package.json
```diff
@@ -75,6 +75,56 @@
...
+        "command": "konveyor.diffView.copyPath",
+        "title": "Copy Path"
+      },
+      {
+        "command": "konveyor.diffView.next",
+        "title": "Go to Next File"
+      },
+      {
+        "command": "konveyor.diffView.prev",
+        "title": "Go to Previous File"
```
djzager, line 127: What I was expecting as an implementation of #70, was to have a new command like `konveyor.viewFix` that would be called by the webview when a user clicks a certain button/target (like the filename). This command would the actual change to be showed to the user in context as either a unified/inline or a split diff. That diff can either be "accept"ed (maybe a `konveyor.acceptFix`) or "reject"ed (`konveyor.rejectFix`).

I reject the idea that our extension needs to follow the example of a SCM.

All I want to see is the ability to see a specific fix in it's context.

djzager (changes requested):
I am a hard no on this PR as it is currently constructed. To reiterate what I have said on Slack, this is what I think the main user flow through our extension should look in mid-November:

1. User installs our extension and is presented with a walkthrough where they can configure the analyzer, configure kai, sources+targets, etc.
2. User starts the server and we initiate the client/server communication with either the analyzer or kai
3. User calls konveyor.runAnalysis where we run an analysis based on configuration, push results to the Problems (as vscode.Diagnostics) and save those RuleSets to our extensions active workspace state.
4. User can view and interact with the results of an analysis in either our webview OR the problems panel. All that cool filter/sort/search, whatever that lets you navigate the results.
5. User selects an individual incident of a particular violation and they call konveyor.getSolution, and we make a call to kai (assuming genAi is enabled) to get a solution. The result we get back will be a bunch of changes that kai wants to make to the repository for that solution.
6. With a specific "solution" returned, the user can navigate (maybe breadcrumbs? and state management so we can come back) the changes to be made. This is where I feel the biggest disconnect. We should primarily be driving this user experience through our singular webview (this is what we can take with us when we go to implement this thing in another editor like jetbrains or eclipse). So we make it possible for the user to "see" the changes to be made in our singular webview and they can accept OR reject them. Additionally, if you click on the code block we will open that specific file in the editor for you and show you that diff inline where the user can accept OR reject a specific change.

This PR deals with a very specific interaction inside of No.6 above, I want a function to be callable from the webview that shows a specific fix in its context (If you want to do accept/reject fix from #72 ... go for it), but the main feature addition in this pull request should be **viewing a specific fix in the editor in context**.

vscode/src/extension.ts
rszwajko: IMHO it fits our roadmap. 

As for `konveyor.diffView.applyAll`,  you are right - the command should be in the global namespace i.e. `konveyor.applyAll`

vscode/package.json
rszwajko: I'll add a `konveyor.viewFix` command (the interface/name may  change).

>That diff can either be "accept"ed (maybe a konveyor.acceptFix) or "reject"ed (konveyor.rejectFix).
>
Note that both the current code (left side of the diff) and resolutions (right side of the diff) might be edited. 
The final code change will include those local changes (once we implement that action).

vscode/src/diffView/register.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+function applyFile(item: FileItem) {
+  item.apply();
+}
+
+function revertFile(item: FileItem) {
+  item.revert();
+}
+async function copyDiff(item: KonveyorFileModel | FileItem | unknown) {
```
djzager, line 66: I am picking this one as an example to support the point I made in [this comment](https://github.com/konveyor/editor-extensions/pull/79/files#r1815412315).

Could you please articulate for me, in detail, why this function could not be defined in [the commandsMap of commands.ts where it would also be registered](https://github.com/konveyor/editor-extensions/blob/65272448356155f4089c64cb8c3dfca2ee68290e/vscode/src/commands.ts#L22C7-L22C18).

rszwajko: @djzager 
Technically there are no problems. However let me extend [the answer above](https://github.com/konveyor/editor-extensions/pull/79#issuecomment-2432592967):
1. all commands are already listed in `package.json` -  note that only string IDs are used so there is no dependency on the code. Command map duplicates this registry BUT
2. differently then `packaged.json` it works on actual code which brings side effects:
   1.   size - the file is already >300 lines and would need to keep growing forever
   2.  imports - the file needs to import code from all modules/units
   3.  volatility - the file is likely to be updated by unrelated commits

If you feel more confident to register commands in one place then I would recommend that each module/unit should export the their small command maps that would get merged before registration.

rszwajko: @djzager 
Note that `command.ts` is the most frequently changed production source code with volatility rate the same as `yarn.lock` . Below 10 most volatile files in Continue's vscode extension (`git log --name-only  --pretty=format: | sort | uniq -c | sort -nr | grep vscode`)
```
    508 extensions/vscode/package.json
    456 extensions/vscode/package-lock.json
    223 extensions/vscode/config_schema.json
    180 extensions/vscode/continue_rc_schema.json
    126 extensions/vscode/src/debugPanel.ts
    118 extensions/vscode/yarn.lock
    118 extensions/vscode/src/commands.ts
    117 extensions/vscode/src/ideProtocol.ts
    103 extensions/vscode/scripts/prepackage.js
     76 extensions/vscode/src/webviewProtocol.ts
```

djzager: It is extremely helpful to me to see these metrics. I think the increased volatility is a fair price to pay to make navigating our codebase at this stage as easy as possible. I'm hopeful that what we discussed in #87 is a reasonable/meaningful compromise that mitigates some of the risk you have identified.

vscode/src/extension.ts
rszwajko: 1. Moved registration into `VsCodeExtension` following the example in https://github.com/konveyor/editor-extensions/pull/90 (by @ibolton336 ).
2. renamed `konveyor.diffView.applyAll` to `konveyor.applyAll` and registered via `command.ts`

vscode/package.json
rszwajko: Added `konveyor.diffView.viewFix` command.

rszwajko:
https://github.com/user-attachments/assets/d3d09e0e-25f6-4781-8682-1eeca3683246

djzager (approved):
LGTM.

PR #83, opened 2024-10-25
:bug: Fix eslint setup after migration to new project layout

rszwajko:


ibolton336 (approved):

rszwajko:
@ibolton336 
it might be my setup but I've encountered multiple eslint problems after the migration.
CLI included hidden vscode folders, the UI was failing to parse project (see below).
I've tweaked the config a bit and it works for me now. Please test on your env.

Screenshot from 2024-10-25 15-13-19

webview-ui/src/App.tsx
```diff
@@ -20,7 +20,7 @@ import { vscode } from "./utils/vscode";
 import GuidedApproachWizard from "./components/GuidedApproachWizard";
 import ProgressIndicator from "./components/ProgressIndicator";
 import ViolationIncidentsList from "./components/ViolationIncidentsList";
-import { Incident, RuleSet } from "./types";
+import { Incident, RuleSet } from "@shared/types";
```
ibolton336, line 23: Oversight on my end removing the tsc compilation step from the webview build script.

rszwajko:
@ibolton336 
If you checkout a fresh project then the problems described above are not present. Not sure what was different in my case ;/ Changes in the eslint config seem neutral.

eslint.config.mjs
```diff
@@ -54,12 +54,7 @@ const tsConfig = {
       ecmaFeatures: {
         jsx: true,
       },
-      project: [
-        "./tsconfig.json",
-        "./shared/tsconfig.json",
-        "./vscode/tsconfig.json",
-        "./webview-ui/tsconfig.json",
-      ],
+      project: true,
```
ibolton336, line 57: Should this be necessary with latest test?

ibolton336 (changes requested):

djzager (approved):

eslint.config.mjs
rszwajko: hmm... after reverting I see the problem again (on my local repo). Only freshly cloned works for me.

rszwajko: I need to use either `project: true` or `projectService: true,`

ibolton336 (dismissed):

eslint.config.mjs
ibolton336: Ah I was able to reproduce this after installing the proper eslint extensions.

ibolton336 (approved):
LGTM after testing the setting on a fedora machine with eslint / eslint-prettier extensions.

PR #84, opened 2024-10-25
:sparkles: Add tsc compile step for webview build

ibolton336:
- Adds `tsc` compilation step for webview. This was accidentally removed. 
  - The tsc output is not required as we are using the vite build output. 
  - Adds types back in for future kai integration work. These were removed in move to new shared package for types.
- Correct the setting "outDir": "out" in the vscode/tsconfig.json file. The compiled JavaScript files from the TypeScript source files will be placed in the vscode/out folder. The .vscode-test.mjs file is then able to load and run the tests from the out/test/**/*.test.js files. Currently tests aren't working. This fixes that. 
- Adds pattern to handle test assets to copy webpack plugin. This will ensure all necessary files are packaged correctly for both local testing and CI/CD environments.
- Removes the need for test compilation step as webpack is taking care of the compilation into the out dir. 

- Adds a "predev" script to make sure the "shared" package is built since both the vscode and webview packages depend on it. This prevents type errors seen when running a dev without a build currently.

# Output Directory Configuration

## Shared Package (`/shared`)
- Output: `./dist`
- Purpose: Contains compiled TypeScript files and type definitions used by other packages
- Configuration:
  - Generates `.d.ts` type definitions and source maps
  - Uses CommonJS modules for broad compatibility
  - Outputs declaration maps for better debugging
  - Files are referenced by both VSCode extension and webview UI

## VSCode Extension (`/vscode`)
- Output: `./out`
- Purpose: Contains the compiled extension code that VSCode loads
- Configuration:
  - Standard VSCode extension output location
  - Uses CommonJS modules as required by VSCode
  - Includes source maps for debugging
  - Bundles the webview UI build output during packaging. Webpack will copy the vite output here (explained in webview section.)

## Webview UI (`/webview-ui`)
- Output: 
  - TypeScript build: `./dist` (type checking only)
  - Vite build: `./build` (production assets)
- Purpose: Dual output strategy
  1. TypeScript compilation (`tsc`) generates type definitions and checks types
  2. Vite builds the actual production assets that get bundled into the extension
- Configuration:
  - Uses ESNext modules for Vite compatibility
  - TypeScript output is not used in production
  - Vite-built assets are copied by the VSCode extension's webpack config during packaging
  - Maintains type safety while leveraging Vite's optimized build process

## Build Process Flow
1. Shared types are built first (`/shared/dist`)
2. Webview UI runs type checking and Vite build
3. VSCode extension builds and packages everything together
4. Final extension package includes:
   - Compiled extension code (`/out`)
   - Webview UI production build
   - Type definitions from shared package


vscode/tsconfig.json
```diff
@@ -2,15 +2,15 @@
...
     "lib": ["ES2020"],
     "sourceMap": true,
     "rootDir": "src",
     "strict": true,
-    "composite": true
+    "composite": true,
+    "declaration": true
   },
-  "include": ["src/**/*.ts"],
-  "references": [{ "path": "../shared" }]
```
rszwajko, line 15: the config worked for me without references but it puzzles me - is it because we only have types in `shared`?

rszwajko (approved):
Works for me!

vscode/tsconfig.json
ibolton336: I agree the references need a second look & perhaps a TODO commend. The way I understand it - since we build the shared package before running dev mode ( or any build / dev of the webview or vscode packages), we should have everything available as the shared package is imported in the package.json for those respective packages

PR #85, opened 2024-10-28
:seedling: Rename x64 kai binaries

sjd78:
The nodejs `os.arch()` function uses the name `x64` and not `amd64` for intel/amd arch. See [1].

[1]: https://nodejs.org/docs/latest-v20.x/api/os.html#osarch


djzager (approved):

djzager:
Thank you @sjd78

ibolton336 (approved):

ibolton336:
image


Ran into this today also - thanks!

PR #86, opened 2024-10-30
:seedling: adding changelog func

savitharaghunathan:
Added changelog functionality

updated the steps to match ci-repo workflow steps

wait for https://github.com/konveyor/release-tools/pull/96 to merge


djzager (approved):

.github/workflows/release.yml
```diff
@@ -58,30 +57,24 @@ jobs:
...
 
       - name: Install dependencies
-        working-directory: ./vscode
         run: npm ci
 
-      # Run tests on Linux
+      # Run tests on different OSes
       - name: Run tests (Linux)
         if: matrix.arch == 'linux'
         working-directory: ./vscode
```
rszwajko, line 65: should we remove also this one? if this is a special case please add a comment

.github/workflows/release.yml
```diff
@@ -58,30 +57,24 @@ jobs:
...
         if: matrix.arch == 'windows'
-        working-directory: ./vscode
         shell: cmd
         run: npm test
 
       - name: Build Package
-        working-directory: ./vscode
         run: npm run package
 
       - name: Generate .vsix package
```
rszwajko, line 80: another case where we leave the `./vscode` folder - if it's a special case please leave a comment

rszwajko (commented):
There are some places left with  `./vscode`  working directory - check my comments below:

.github/workflows/release.yml
sjd78: Yeah it does look like this working-directory line should be removed.

.github/workflows/release.yml
sjd78: This is a repeat from the `ci-repo.yml`.  I don't mind a comment not being here mostly because the packaging script is due some updating.

sjd78 (changes requested):
working directory for the linux testing probably isn't needed

.github/workflows/release.yml
rszwajko: OK. I just got confused that working folder is left only in some places.

sjd78 (approved):
LGTM

PR #87, opened 2024-10-31
:sparkles: Load analysis results from file

rszwajko:
Use case:
1. trigger "Load results from files" command from the palette
2. select analysis results and/or solutions 
    a) file format: JSON 
    b) loading solutions is a stub - to be continued in follow-up 
    c) if multiple files with analysis results(or solutions) are selected only the first one will be used - possible other   approaches are failing the import or merging the files
4. results are loaded into UI - both VS Code views and our webview

Related refactorings:
1. decouple loading the data from analyzer client - now it operates on the state only
2. move diagnosticCollection to the state - it has an extension-wide scope
3. disable triggering "analysisComplete" message from analyzer client - it duplicates "loadStoredAnalysis" message
5. move anaylzerResultsts helper to newly created data module - after loading results functionality was moved from the analyzer client it was not used in this module
6. removed clearing rule sets from "konveyor.runAnalysis" command - delay the action until results will be ready. Loading the new results should overwrite old results.


rszwajko:
https://github.com/user-attachments/assets/3e0bf84e-1edd-4153-877d-867edd4e611c

vscode/src/commands.ts
```diff
@@ -307,7 +296,11 @@ const commandsMap: (state: ExtensionState) => {
 };
 
 export function registerAllCommands(state: ExtensionState) {
-  for (const [command, callback] of Object.entries(commandsMap(state))) {
+  const allCommands = {
+    ...commandsMap(state),
+    ...loadResultsCommands(state),
```
djzager, line 301: For me, navigating this codebase is very important. Having to grep to find exactly where  specific commands are defined is just not fun. I understand that you do not agree for reasons stated [here](https://github.com/konveyor/editor-extensions/pull/79#discussion_r1816413365).

**I propose a compromise**. How about, instead of defining these maps elsewhere and joining them as we are here, we do like they do in [continue commands.ts](https://github.com/continuedev/continue/blob/76877df16e43eabe9bc7d6d7219a34dfc2d75610/extensions/vscode/src/commands.ts#L273C1-L280C7):

```
    "continue.acceptVerticalDiffBlock": (filepath?: string, index?: number) => {
      captureCommandTelemetry("acceptVerticalDiffBlock");
      verticalDiffManager.acceptRejectVerticalDiffBlock(true, filepath, index);
    },
    "continue.rejectVerticalDiffBlock": (filepath?: string, index?: number) => {
      captureCommandTelemetry("rejectVerticalDiffBlock");
      verticalDiffManager.acceptRejectVerticalDiffBlock(false, filepath, index);
    },
```

This way, as a developer, when I go to look for the implementation of a command I find in the palette, I find it by it's display name in the package.json, I come here to find it's implementation...then the level to which it is actually implemented in this file is up to you.

vscode/package.json
```diff
@@ -68,6 +68,21 @@
...
+      },
+      {
+        "command": "konveyor.cleanRuleSets",
+        "title": "Reset analysis results",
+        "category": "Konveyor"
+      },
+      {
+        "command": "konveyor.loadRuleSets",
+        "title": "Load analysis results",
+        "category": "Konveyor"
```
djzager, line 85: I'm curious if these are something we only see being useful for our development workflows or useful to end users as well. My suspicion is that it best serves us in development.

We do not have to include commands in this file in order to call them like `vscode.commands.executeCommand("konveyor.loadRuleSets", rulesets);`, only register them ([ref](https://stackoverflow.com/a/55271392)).

For example, maybe we omit them from here and have a "development" panel in our webview to access some of these commands that we need for development purposes.

djzager (changes requested):
This is awesome. Thank you for putting this together. The only issues I see are related to commands.ts and whether or not these should be in our command palette.

I am nearly certain the commands.ts commentary is annoying, but I do hope that I proposed a reasonable compromise.

vscode/src/commands.ts
rszwajko: Compromise accepted:) This will reduce volatility rate. FYI check the [volatility stats](https://github.com/konveyor/editor-extensions/pull/79#discussion_r1825138624) for Continue's vscode extension.

vscode/package.json
djzager: Fair enough. I think we should maybe consider using the when clause to limit when some of these palette commands are available.

https://code.visualstudio.com/api/references/when-clause-contexts

We should just create an issue so we can prioritize this.

djzager (approved):
Visually this all lgtm. I'm sure I'll be using this shortly.

vscode/src/test/analyzerResults.test.ts
```diff
@@ -1,27 +1,24 @@
...
   test("processIncidents should populate diagnostics correctly", () => {
     const filePath = path.resolve(__dirname, "./testData/output-data.yaml");
     const ruleSets: RuleSet[] | undefined = readYamlFile(filePath);
     assert.ok(ruleSets, "RuleSets should be loaded from YAML file");
-    const diagnosticsMap = new Map<string, vscode.Diagnostic[]>();
-    processIncidents(ruleSets!, diagnosticsMap);
+    const results = processIncidents(ruleSets!);
     assert.ok(
-      diagnosticsMap.has("file:///opt/input/source/src/main/webapp/WEB-INF/web.xml"),
+      results.every(([uri]) => uri.fsPath === "/opt/input/source/src/main/webapp/WEB-INF/web.xml"),
```
ibolton336, line 14: Wonder if we need to normalize the path here to be platform agnostic? To satisfy CI we could try path.posix and path.normalize.

rszwajko: good catch @ibolton336 ! It was the path problem. After I've refactored the test the assertion failure was clear:
```
   AssertionError [ERR_ASSERTION]: web.xml should have 4 diagnostics
      + expected - actual

       [
...
      -  "\\opt\\input\\source\\src\\main\\webapp\\WEB-INF\\web.xml"  
      +  "/opt/input/source/src/main/webapp/WEB-INF/web.xml"
...
       ]
```
`normalize()` did'nt work but split on `path.sep` did.

vscode/package.json
rszwajko: IMHO it's low priority. Adding `when` condition will basically limit the command visibility to some context i.e. user will need to open our panels to see the command. Once we reach maturity it may be an elegant way of saying that the command is internal/debug/dev etc.

ibolton336 (approved):

djzager (approved):
nice

PR #90, opened 2024-11-05
:sparkles: Integrate results view with diff view & add Kai fix webview panel 

ibolton336:
- Add webview messaging for solution request and confirmation for communication between all 3 webviews. New handlers created. 
- When a solution is requested (getSolution command), the solution is immediately applied.
  - This triggers the opening of a kai fix viewer (webview panel) that displays the violation and other incidents based on lightbulb click context. 
  - Sets up the plumbing to allow the creation of a new webview that supports opening a panel. Augmented the current webview provider to allow this.
  - This new panel is initialized on solution request. Resolution panel outer webview is created on startup just like the analysis webview -  Was able to setup a new var for view type and pass that down when webview is initialized so that our react app can conditionally render / route to different components based on this var. So we have 2 entirely seperate webviews that can be messaged and handle requests .. we just have to grab it from the new map in extension state and make a request to it. They both are fed the same react app, we just conditionally show what is there based on the viewtype var.

- Logic to handle queue for resolution panel since it is created & messaged at same time during solution request. 
  - Opening a webview panel vs a webview has its own lifecycle that made creating the panel & then sending a message to it very interesting.
The panel needs to exist and have the listener set up before it can receive messages, so needed to implement a queue system. Messages are pushed to the queue if the webview panel isnt ready and then looped through and sent once the panel exists.
- Refactor the analysis view card to use PF6 card header & expansion techniques. Allows us the ability to use the dropdown as a part of the card actions. This is where the fix all button currently sits which requests solutions for all incidents within the parent violation umbrella. 
- Adds a lightbulb icon at the incident level for triggering single incident solution requests. 

TODO:
-  Expand the sharedState in extension state to handle state flags for isAnalyzing and isWaiting for solution. Make use of event emitters. https://github.com/konveyor/editor-extensions/issues/98 
- Determine how webview reacts to analysis run from command palette directly and ensure no gaps https://github.com/konveyor/editor-extensions/issues/99


webview-ui/src/App.tsx
```diff
@@ -190,18 +195,25 @@ const App: React.FC = () => {
...
                 focusedIncident={focusedIncident}
                 onIncidentSelect={handleIncidentSelect}
-                onGetSolution={() => console.log("Get Solution")}
-                onGetAllSolutions={(violation) => {
-                  console.log("Get All Solutions", violation);
-                  // vscode.postMessage({
-                  //   command: "fixAll",
-                  //   violation,
-                  // });
+                onGetSolution={(selectedViolation: Violation) => {
```
rszwajko, line 198: can we use here the command directly i.e. `vscode.commands.executeCommand(....)` ?

vscode/src/webviewMessageHandler.ts
```diff
@@ -4,6 +4,18 @@ import { ExtensionState } from "./extensionState";
 export function setupWebviewMessageListener(webview: vscode.Webview, _state: ExtensionState) {
   webview.onDidReceiveMessage(async (message) => {
     switch (message.command) {
+      case "getSolution": {
+        vscode.commands.executeCommand("konveyor.getSolution");
```
rszwajko, line 8: the commands are async - do we intent do run them sequentially ?

vscode/src/diffView/register.ts
```diff
@@ -39,30 +41,46 @@ export function registerDiffView(context: vscode.ExtensionContext): void {
...
+
     vscode.commands.registerCommand("konveyor.diffView.copyDiff", copyDiff),
     vscode.commands.registerCommand("konveyor.diffView.copyPath", copyPath),
   );
 }
 
-function applyAll() {
+function applyAll(state: ExtensionState) {
   vscode.window.showInformationMessage(`[TODO] Apply all resolutions`);
+  state?.sidebarProvider?.webview?.postMessage({ type: "solutionConfirmation" });
```
rszwajko, line 62: this part will be refactored heavily - if possible it's better to defer changes until https://github.com/konveyor/editor-extensions/pull/79 is merged

webview-ui/src/App.tsx
```diff
@@ -14,26 +14,25 @@ import {
...
   const [isAnalyzing, setIsAnalyzing] = useState(false);
   const [analysisMessage, setAnalysisMessage] = useState("");
   const [errorMessage, setErrorMessage] = useState<string | null>(null);
-  const [isWizardOpen, setIsWizardOpen] = useState(false);
   const [focusedIncident, setFocusedIncident] = useState<Incident | null>(null);
   const [expandedViolations, setExpandedViolations] = useState<Set<string>>(
     new Set(),
   );
-  const [isChatVisible, setIsChatVisible] = useState(false);
+  const [isWaitingForSolution, setIsWaitingForSolution] = useState(false);
```
rszwajko, line 35: I'd rather test for the data itself:
1. there is a risk that the flag will be out-of-sync with data
2. this will complicate mock commands

rszwajko:
@ibolton336 
check newest changes in https://github.com/konveyor/editor-extensions/pull/79 - I was finally able to integrate with `VsCodeExtension` and real(loaded) solutions.

vscode/src/KonveyorGUIWebviewViewProvider.ts
```diff
@@ -4,47 +4,92 @@ import {
...
+          this._isWebviewReady = false; // Reset readiness flag on disposal
+        },
+        null,
+        this._disposables,
+      );
+    }
+  }
 
-    webviewView.webview.options = {
+  private initializeWebview(webview: Webview): void {
```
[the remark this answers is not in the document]
ibolton336: It is called for webview creation on extension initial load and on panel creation for the resolutions panel.

webview-ui/src/components/AnalysisPage.tsx
```diff
@@ -0,0 +1,206 @@
...
+import spacing from "@patternfly/react-styles/css/utilities/Spacing/spacing";
+
+import { vscode } from "../utils/vscode";
+import ProgressIndicator from "./ProgressIndicator";
+import ViolationIncidentsList from "./ViolationIncidentsList";
+import { Incident, RuleSet, Violation } from "@shared/types";
+
+const AnalysisPage: React.FC = () => {
+  const [analysisResults, setAnalysisResults] = useState<RuleSet[] | null>();
+  const [isAnalyzing, setIsAnalyzing] = useState(false);
```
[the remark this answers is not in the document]
ibolton336: +1 we should leverage the event emitter and shared state to keep things in sync better vs using react state here. Will open an issue.

vscode/src/webviewMessageHandler.ts
```diff
@@ -1,9 +1,111 @@
...
+              diff: "diff --git a/src/main/java/com/redhat/coolstore/service/CatalogService.java b/src/main/java/com/redhat/coolstore/service/CatalogService.java\nindex 422a3f4..9a6feff 100644\n--- a/src/main/java/com/redhat/coolstore/service/CatalogService.java\n+++ b/src/main/java/com/redhat/coolstore/service/CatalogService.java\n@@ -9,12 +9,12 @@ import javax.persistence.criteria.CriteriaBuilder;\n import javax.persistence.criteria.CriteriaQuery;\n import javax.persistence.criteria.Root;\n \n-import javax.ejb.Stateless;\n+import jakarta.enterprise.context.ApplicationScoped;\n import javax.persistence.EntityManager;\n \n import com.redhat.coolstore.model.*;\n \n-@Stateless\n+@ApplicationScoped\n public class CatalogService {\n \n     @Inject\n",
+            },
+          ],
+        };
+
+        vscode.commands.executeCommand("konveyor.diffView.focus");
+        vscode.commands.executeCommand("konveyor.showResolutionPanel");
+        const incidentUri = incident.uri;
+
+        // Remove 'file://' from the URI and ensure it doesn't start with a leading slash
```
rszwajko, line 30: if `uri` is a valid string representation of `vscode.Uri` then we could try `Uri.parse(uri)` and use it as in other places i.e. `fromRelativeToKonveyor(vscode.workspace.asRelativePath(uri)),`

vscode/src/webviewMessageHandler.ts
```diff
@@ -1,9 +1,111 @@
...
 
-export function setupWebviewMessageListener(webview: vscode.Webview, _state: ExtensionState) {
+export function setupWebviewMessageListener(webview: vscode.Webview, state: ExtensionState) {
   webview.onDidReceiveMessage(async (message) => {
     switch (message.command) {
+      case "getSolution": {
+        const { violation, incident } = message;
+        const [analysisResults, solution] = await loadStateFromDataFolder();
+
+        const mockSolution: GetSolutionResult = {
```
rszwajko, line 15: I guess we can remove hardcoded mocks now.

vscode/src/webviewMessageHandler.ts
```diff
@@ -1,9 +1,111 @@
...
+        const incidentKonveyorUri = fromRelativeToKonveyor(
+          vscode.workspace.asRelativePath(incidentPath),
+        ); // Convert to konveyor:// format
+
+        console.log("incidentKonveyorUri", incidentKonveyorUri);
+        console.log("incidentPath", incidentPath);
+        console.log("solution", solution);
+
+        // Step 2: Check if any of the solution changes are relevant to the incident
+        const isRelevantSolution = solution?.changes.some((change: Change) => {
```
rszwajko, line 47: IMHO we should accept whatever the AI in the backend will produce. This should also make our life easier during development (mock data are not covering all cases).

vscode/src/webviewMessageHandler.ts
```diff
@@ -1,9 +1,111 @@
...
+
+        break;
+      }
+      case "getAllSolutions": {
+        // vscode.commands.executeCommand("konveyor.getAllSolutions");
+        vscode.commands.executeCommand("konveyor.diffView.focus");
+        break;
+      }
+      case "viewFix": {
+        const { change, incident } = message;
```
rszwajko, line 75: it's not always incident URI - I would change the naming to more generic one.

rszwajko (approved):
Looks OK! We can discuss the points in comments later on and if needed fix in follow-up. It would be nice to attach a screencast or screenshots of the the full flow.

rszwajko (approved):

ibolton336:
https://github.com/user-attachments/assets/8c6af716-64d1-4272-a576-38a199f42d4c

https://github.com/user-attachments/assets/f7e73094-dea8-4d14-af3c-9d38a6edc5ac

PR #91, opened 2024-11-05
:sparkles: Load solutions from file

rszwajko:
Changes:
1. add konveyor.loadSolution command used by existing command konveyor.loadStaticResults
2. store loaded analysis results and received solution inside ".vscode/konveyor" folder
    a) last 5 files are stored for each type 
    b) file name is unique - with human readable sufix (datetime)
4. change ECMA level to "ES2021" for libs to get String.replaceAll()
5. introduce Change interface with props:
    a) "original" - relative file path before the change, may be empty if file was created in this change 
    b) "modified" - relative file path after the change, may be empty if file was deleted in this change 
    c) "diff" - diff in unified format - tested with git diffs

### DIFF
--- a/mocks/coolstore_fixEverywhere_javax_ejb_Stateless.json
+++ b/mocks/coolstore_fixEverywhere_javax_ejb_Stateless.json
@@ -1,0 +1,35 @@
+{
+    "errors": [],
+    "changes": [
+        {
+            "original": "pom.xml",
+            "modified": "pom.xml",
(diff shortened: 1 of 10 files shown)


rszwajko:
### Screencast

https://github.com/user-attachments/assets/0cd7638e-389e-4787-bed0-fdd86424dd88

ibolton336:
Looking good just need to add the "mocks" directory to the .vscodeIgnore to satisfy the package CI run.

djzager (approved):

PR #100, opened 2024-11-15
:seedling: Refactor shared package imports

sjd78:
The `@shared/types` package does not actually exist. It was only working because there was a vestigial reference for it in the `package-lock.json` file from before the shared workspace package was renamed from `@shared/types` to `@editor-extensions/shared`. `@shared/types` was pointing to the shared workspace.

Fixes:
  - Remove the `@shared/types` reference in the lock file
  - Refactor all of the imports to pull from the current npm package name

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1518,10 +1518,6 @@
         "win32"
       ]
     },
-    "node_modules/@shared/types": {
-      "resolved": "shared",
-      "link": true
-    },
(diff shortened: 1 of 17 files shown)


djzager (approved):
ACK

ibolton336:
Good catch!

ibolton336 (approved):

PR #101, opened 2024-11-15
:seedling: Simplify and harmonize linting and dependencies

sjd78:
Summary
-------

Looking at the existing package scripts, there is opportunity to harmonize how they are defined and used. They've been updated as such, and to natively use workspace configurations.

Changes
-------

- Force npm adherence to root `package.json` engine settings

- Rearrange `.editorconfig` files
  - moved the `vscode/.editorconfig` to root to apply to the entire repo
  - `docs/.editorconfig` only needs print width, all others inherit from the root

- Changes to root `package.json`:
  - Reorder the workspaces to be in build order so the `--workspaces` commands works nice

  - Add scripts `clean`, `clean:all` and `postinstall`

  - Simplify some linting:
    - Update the "lint" and "build" root scripts to use `--workspaces` to keep things workspace consistent

    - Update root level `lint-staged` settings to generally format any file that prettier knows how to format in any directory that does not have its own `lint-staged` settings

    - Update root `.prettierrc.mjs` to explicitly set the common settings we care about to make sure prettier defaults don't change anything major on an upgrade

- Changes to `shared/package.json`:
  - Drop `devDependencies` that are inherited from the root `package.json`
  - Add scripts `lint` and `lint:fix`
  - Add `lint-staged` settings for the project

- Changes to `vscode/package.json`:
  - Drop dependencies that are inherited from the root `package.json` or are no longer needed
  - Add scripts `lint` and `lint:fix`
  - Add `lint-staged` settings for the project

- Changes to `webview-ui/package.json`:
  - Add scripts `lint` and `lint:fix`
  - Add `lint-staged` settings for the project


djzager (dismissed):
ACK

rszwajko (approved):
Works for me!

djzager (approved):
Thank you for taking care of this.

PR #103, opened 2024-11-19
:sparkles: Support reviewing resolutions with merge editor

rszwajko:
Changes:
1. add optional configuration property "diffEditorType". If set to
   "merge" then open resolutions using merge editor. Otherwise use
   the diff editor.
2. create text document content provider for serving read only virtual
   files to be used as left side of the merge. Register the provider
   under "konveyorReadOnly" scheme.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -163,10 +163,22 @@
         "title": "Apply Block"
       },
       {
+        "command": "konveyor.diffView.applySelectionInline",
+        "icon": "$(check)",
+        "category": "diffEditor",
+        "title": "Apply Selection"
+      },
+      {
         "command": "konveyor.diffView.applyBlock",
         "icon": "$(arrow-left)",
         "category": "diffEditor",
         "title": "Apply Block"
+      },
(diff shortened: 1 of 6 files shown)


ibolton336:
https://github.com/user-attachments/assets/8fc161da-088b-4a11-947a-72835217172e

Noticing that the diffview isnt reflecting the resolved changes in merge mode. Should the list of resolutions be updated once all of the changes are resolved in the editor panel?

rszwajko:
@ibolton336 
>Noticing that the diffview isnt reflecting the resolved changes in merge mode. Should the list of resolutions be updated once all of the changes are resolved in the editor panel?
>
The actions used there are built into VS Code - they are not aware that it's a Konveyor flow so we aren't notified.  Note that at this level "accepting" a solution is not well defined. User can accept some lines, ignore others and add his  own modifications. We could try however to add some logic based on resolved conflicts or save actions to check that mark file as "processed".

ibolton336 (approved):

PR #104, opened 2024-11-19
:bug:  Reset Path

hhpatel14:
Previously, when you selected a file to override the default path, the path was saved and couldn't be reset. With this update:

- If you open the file selection dialog but don’t choose a file, the path will reset to undefined.
- When no path is set, the default binary from the /assets/bin folder will be used.


djzager (approved):
lgtm

PR #107, opened 2024-11-20
:sparkles: Use immer for extension state

rszwajko:
Depends on #103 

Concentrate data state in the extension object as state.data and enforce top-down propagation of changes.
Key points:
1. state is made immutable using Immer - reference comparison can be used to detect changes
2. components interested in receiving notifications about state change should add callbacks via onDidChangeData
3. by convention after initialization the state is modified only via commands 

Related refactorings:
1. track per-file solution state - LocalChange has now "state" prop with 3 values: pending,applied,discarded
2.  use onDidChangeData to trigger file model actions (content of the native Konveyor Resolution view) and in-memory FS actions (suggestion diffs applied to current code)


sjd78:
I forget how clean the producers make changing the data look!

ibolton336 (approved):

PR #109, opened 2024-11-21
:sparkles: Add retry and error handling to server start for webview

ibolton336:
https://github.com/user-attachments/assets/6d311974-794f-4940-a700-84ac7b3ac87c


webview-ui/src/components/AnalysisPage.tsx
```diff
@@ -30,6 +30,8 @@ const AnalysisPage: React.FC = () => {
   const [expandedViolations, setExpandedViolations] = useState<Set<string>>(new Set());
   const [isWaitingForSolution, setIsWaitingForSolution] = useState(false);
   const [serverRunning, setServerRunning] = useState(false);
+  const [serverStarting, setServerStarting] = useState(false);
```
rszwajko, line 33: Keeping the flags in sync is tedious - how about deriving them from state directly as language client - check [`isRunning` helper.](https://github.com/microsoft/vscode-languageserver-node/blob/8e6bf18f22b2e83b60a27a51447b99acf98c4381/client/src/common/client.ts#L1236)

vscode/src/client/analyzerClient.ts
```diff
@@ -5,47 +5,81 @@ import * as fs from "fs";
...
-      if (err) {
-        vscode.window.showErrorMessage("Java is not installed. Please install it to continue.");
-        return;
-      }
-    });
-    exec("mvn -version", (err) => {
-      if (err) {
-        vscode.window.showErrorMessage("Maven is not installed. Please install it to continue.");
+
+    exec("java -version", (javaErr) => {
```
rszwajko, line 31: future improvement - running those commands take some time - we could run them AFTER the server failed and/or run them once (and store the flag even in `globalState`).

PR #110, opened 2024-11-22
:seedling: Create the vsix package from `dist/`

sjd78:
Explicitly copy all files needed by the vscode extension to a `dist/` directory and build the `.vsix` file from there.

There are a few advantages to using this method:

  - Only what we explicitly copy to the `dist/` directory will be included in the `.vsix` file

  - Every developer and CI script will be able to use the same packaging scripts to maintain consistency

  - Changes to what is included is centralized in a few scripts

Note: In hopes of easier build support on windows, the scripts are all javascript/nodejs.  Getting bash scripts to be stable
on windows can be a challenge, and maintaining platform specific scripts is best avoided if possible.

### DIFF
--- a/.github/workflows/ci-repo.yml
+++ b/.github/workflows/ci-repo.yml
@@ -65,7 +65,12 @@ jobs:
           npm ci
 
       - name: Lint sources
-        run: npm run lint
+        run: |
+          npm run lint
+
+      - name: Build
+        run: |
+          npm run build
 
   test:
     name: Test (${{ matrix.arch }})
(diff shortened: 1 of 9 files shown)


rszwajko (commented):
Works for me!
I've noticed 2 issues:
1. `npm run package` just fails when run alone - it would be nice to add some comments or execute the dependencies
2. webview is missing files - we might have some hardcoded paths there
Screenshot from 2024-11-25 15-00-59

ibolton336:
> Works for me! I've noticed 2 issues:
> 
> 1. `npm run package` just fails when run alone - it would be nice to add some comments or execute the dependencies
> 2. webview is missing files - we might have some hardcoded paths there
>    Screenshot from 2024-11-25 15-00-59
> 
> @ibolton336 ^^ it seems we need to fix paths here ->
> 
> https://github.com/konveyor/editor-extensions/blob/ae7f628e0200ce67461b08f8a3ecd319783b20b7/vscode/src/KonveyorGUIWebviewViewProvider.ts#L107

Fixed here: 

https://github.com/konveyor/editor-extensions/pull/113

savitharaghunathan (approved):
/lgtm
works for me

rszwajko (approved):

PR #115, opened 2024-11-25
:seedling: Update release workflow with new npm targets

savitharaghunathan:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -72,6 +72,10 @@ jobs:
         shell: cmd
         run: npm test
 
+      # clean
+      - name: Clean build
+        run: npm run clean
+
       # Collect assets
       - name: Collect assets
         run: npm run collect-assets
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -73,20 +72,24 @@ jobs:
         shell: cmd
         run: npm test
 
-      - name: Build Package
-        run: npm run package
+      # Collect assets
```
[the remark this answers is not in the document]
savitharaghunathan: added a step to run clean. PTAL

.github/workflows/release.yml
```diff
@@ -73,20 +72,28 @@ jobs:
         shell: cmd
         run: npm test
 
-      - name: Build Package
-        run: npm run package
+      # clean
+      - name: Clean build
+        run: npm run clean
```
sjd78, line 77:
```suggestion
        run: |
          npm run clean
          npm run build
```

sjd78 (changes requested):
One more changes and it'll be good to go.

.github/workflows/release.yml
savitharaghunathan: thank you. added build to the step

PR #121, opened 2024-11-26
:seedling: Split `collect-assets.js` code, add `tar.gz` download

sjd78:
Split the contents of `collect-assets.js` into two files. The base file has the release details.  The new `_download.js` file has the functions that do the download and extract work.

This split helps keep the `collect-assets.js` script as small as possible to make future configuration changes easier.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,6 +46,7 @@
         "lint-staged": "^15.2.10",
         "prettier": "^3.3.3",
         "rimraf": "^6.0.1",
+        "tar": "^7.4.3",
         "typescript": "^5.6.3",
         "typescript-eslint": "^8.14.0",
         "unzipper": "^0.12.3"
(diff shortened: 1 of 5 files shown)


savitharaghunathan:
@sjd78 Thanks for the refactor. should we have a config or json file that says all the details of the repo we want to download? The idea is to not touch the scripts but if it is over engineering for what we are trying to achieve please ignore my comment.
I am thinking how to make that as configurable from the release pipeline. say if we want to release two vsix pacakage with different versions of kai, we can just change the tag for kai. 

for now, what we have should work. I am just thinking long term and dont have a good solution right now

sjd78:
> @sjd78 Thanks for the refactor. should we have a config or json file that says all the details of the repo we want to download? The idea is to not touch the scripts but if it is over engineering for what we are trying to achieve please ignore my comment. I am thinking how to make that as configurable from the release pipeline. say if we want to release two vsix pacakage with different versions of kai, we can just change the tag for kai.
> 
> for now, what we have should work. I am just thinking long term and dont have a good solution right now

With these changes, the `collect-assets.js` file is essentially just a config file.  All of the code that does the work of finding, downloading, and extracting are in the other helper files.

Going forward:
  - `collect-assets.js` will say what to collect and where to put it
  - `copy-dist.js` will say what to put in `dist/` from repo sources and collected assets for the vsix

savitharaghunathan (approved):
/lgtm

tested it my machine and it works. thanks @sjd78

rszwajko (approved):
LGTM

PR #123, opened 2024-11-28
:sparkles: Retrieve the list of modified files from the diff

rszwajko:
Map solution response to internal format using only information from the provided diff.

Main advantages:
1. automatically map files to corresponding diffs - apply/discard operations are file based and require single-file diff
2. allows incremental update to support add/delete/rename
3. file paths in git diffs are relative to repository root which usually maps to VS Code workspace

Main disadvantage is increasing the dependency on the git diff format.

Reference-Url: https://github.com/konveyor/kai/issues/502

### DIFF
--- a/shared/src/types/index.ts
+++ b/shared/src/types/index.ts
@@ -94,7 +94,7 @@ export interface LocalChange {
 
 export interface ResolutionMessage {
   type: string;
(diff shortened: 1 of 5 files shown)


djzager (approved):
VISACK

PR #124, opened 2024-11-29
:sparkles:  Reintroduce multi-file support

rszwajko:
Refactor and where necessary partially revert the code introduced in feature/kai branch.

Changes:
1. base the webview view/apply/discard actions on state property
   localChanges which contains parsed and localized resolutions.
2. refactor async code in analyzer client
   a) use promisfied versions of exec and setTimeout
   b) create synchronous version of  canAnalyze
   c) extract extension state mutations to dedicated fire*Change methods
   d) introduce ServerState enum
3. save raw response (solutionData) and scope using loadSolution command
   to better link those entities.
4. re-create Uri objects received in the message handler from webviews.
   After crossing the iframe boundary only data props are preserved.
5. remove triggering checkServerStatus message from the webviews as the
   new onDataChanged event should provide the notifications already
6. derive the list of unprocessed resolution from LocalChange.state
7. create Scope interface - to be extended with more combinations in the
   future.
8. add scope property to Solution and allow viewing incidents without
   running server. Together this allows end-to-end testing with mocks.
9. name error property in GetSolutionResult(ideal response) for better
   alignment with SolutionResponse(actual response)

### DIFF
--- a/webview-ui/src/components/AnalysisPage.tsx
+++ b/webview-ui/src/components/AnalysisPage.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useMemo, useEffect } from "react";
+import React, { useState, useMemo } from "react";
 import {
   Button,
   ButtonVariant,
@@ -49,16 +49,6 @@ const AnalysisPage: React.FC = () => {
       line: incident.lineNumber,
     });
   };
-
-  const fetchServerStatus = () => {
-    vscode.postMessage({ command: "checkServerStatus" });
-  };
-
-  useEffect(() => {
-    fetchServerStatus();
-    const interval = setInterval(fetchServerStatus, 5000);
-    return () => clearInterval(interval);
-  }, []);
 
   const messageHandler = (message: any) => {
     console.log("Received message from VS Code:", message);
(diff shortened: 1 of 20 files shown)


rszwajko:
https://github.com/user-attachments/assets/bb3c4ab3-560c-4f83-aa4d-c9f97b802044

webview-ui/src/components/AnalysisPage.tsx
```diff
@@ -119,7 +109,7 @@ const AnalysisPage: React.FC = () => {
     );
   }
 
-  if (!serverRunning) {
+  if (!serverRunning && !hasViolations) {
```
ibolton336, line 112: Is this just for local development purposes?

ibolton336 (changes requested):

ibolton336 (approved):

webview-ui/src/components/AnalysisPage.tsx
ibolton336: I think we should add a follow up PR to disable server dependent actions like getSolution / runAnalysis if the server is not running while keeping this "new" use case in consideration.

PR #125, opened 2024-12-02
:sparkles: Implement Flux-style message passing

rszwajko:
Changes:
1. convert ServerState enum to union type - enums are not native JS constructs and are not correctly packaged in the current build
2. unify data flow from extension to the webview - the entire state object is sent to keep the state atomic
3. provide type safe action creators (Redux-style)
4. provide generic map-based message handler similar to Redux reducers


webview-ui/src/hooks/useExtensionState.ts
```diff
@@ -0,0 +1,34 @@
+import { useEffect, useState } from "react";
```
ibolton336, line 1: Nice!

ibolton336 (approved):

PR #126, opened 2024-12-03
:sparkles: Add kai configurables to extension config

djzager:
This PR adds the kai configuration items to our extension's config **AND** attempts to standardize the way we get and update our configuration items.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -456,16 +456,15 @@
           "description": "URL for the Kai solution server",
           "scope": "window"
         },
-        "konveyor.kai.providerModel": {
-          "type": "string",
-          "default": "meta-llama/llama-3-1-70b-instruct",
-          "description": "URL for the Kai solution server",
-          "scope": "window"
-        },
-        "konveyor.kai.providerParameters": {
+        "konveyor.kai.providerArgs": {
           "type": "object",
-          "default": {},
-          "description": "Provider parameters",
+          "default": {
+            "model_id": "meta-llama/llama-3-1-70b-instruct",
+            "parameters": {
+              "max_new_tokens": 2048
+            }
+          },
+          "description": "Kai provider arguments",
           "scope": "window"
         }
       }
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -363,37 +363,44 @@
           "default": "",
           "description": "Path to the rpc-server binary. If not set, the extension will use the bundled binary.",
           "scope": "machine",
-          "order": 0
+          "order": 1
         },
-        "konveyor.incidentLimit": {
+        "konveyor.logLevel": {
+          "type": "string",
+          "default": "debug",
+          "description": "Log level to use with extension",
+          "scope": "window",
+          "order": 2
+        },
+        "konveyor.analysis.incidentLimit": {
           "type": "number",
           "default": 10000,
           "description": "Maximum number of incidents to report",
           "scope": "window",
           "order": 1
         },
-        "konveyor.contextLines": {
+        "konveyor.analysis.contextLines": {
           "type": "number",
           "default": 10,
           "description": "Number of lines of context to include in incident reports",
           "scope": "window",
           "order": 2
         },
-        "konveyor.codeSnipLimit": {
+        "konveyor.analysis.codeSnipLimit": {
           "type": "number",
           "default": 10,
           "description": "Number of lines of code to include in incident reports",
           "scope": "window",
           "order": 3
         },
-        "konveyor.useDefaultRulesets": {
+        "konveyor.analysis.useDefaultRulesets": {
           "type": "boolean",
           "default": true,
           "description": "Whether analysis should use the default rulesets, included in the extension",
           "scope": "window",
           "order": 4
         },
-        "konveyor.customRules": {
+        "konveyor.analysis.customRules": {
           "type": "array",
           "items": {
             "type": "string"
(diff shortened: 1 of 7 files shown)


vscode/src/client/analyzerClient.ts
```diff
@@ -106,21 +120,19 @@ export class AnalyzerClient {
...
-          //   max_new_tokens: 2048,
-          // },
+          model: getConfigKaiProviderModel(),
+          parameters: getConfigKaiProviderParameters(),
         },
       },
-      file_log_level: "debug",
-      demo_mode: true,
+      file_log_level: getConfigLogLevel(),
+      demo_mode: Extension.getInstance(this.extContext).isProductionMode ? "false" : "true",
```
rszwajko, line 135: can we make it a hidden config property?  `true | false | devMode (default)`

vscode/src/utilities/configuration.ts
```diff
@@ -0,0 +1,150 @@
...
+
+export async function updateContextLines(value: number): Promise<void> {
+  await updateConfigValue("analysis.contextLines", value, vscode.ConfigurationTarget.Workspace);
+}
+
+export async function updateCodeSnipLimit(value: number): Promise<void> {
+  await updateConfigValue("analysis.codeSnipLimit", value, vscode.ConfigurationTarget.Workspace);
+}
+
+export async function updateUseDefaultRulesets(value: boolean): Promise<void> {
```
rszwajko, line 108: In the current naming scheme rule sets are joined using camel case ( `ruleSets`). Spell-checker reports that as an error which adds extra noise.
image

rszwajko (dismissed):
Works for me! 2 minor comments (see below)

sjd78 (approved):
The config changes look good to me.

I didn't realize the kai rpc server needs a path to the analyzer in the initializeParams.  Good to know!

vscode/src/client/analyzerClient.ts
```diff
@@ -412,39 +430,18 @@ export class AnalyzerClient {
...
-  }
-
-  public getContextLines(): number {
-    return this.config!.get("contextLines") as number;
-  }
-
-  public getCodeSnipLimit(): number {
-    return this.config!.get("codeSnipLimit") as number;
-  }
+  public getRules(): string[] {
```
ibolton336, line 433: Returning an array seems to create an issue [here](https://github.com/konveyor/editor-extensions/pull/126/files#diff-7bf6d907ee218343f5466cc2b00dd02fb1e52c9e918671e641a83fb1a8295d6aL139)

ibolton336 (changes requested):

vscode/src/client/analyzerClient.ts
```diff
@@ -106,21 +130,19 @@ export class AnalyzerClient {
...
       model_provider: {
-        provider: "ChatOpenAI",
+        provider: getConfigKaiProviderName(),
         args: {
-          model: "gpt-3.5-turbo",
-          // parameters: {
-          //   max_new_tokens: 2048,
-          // },
+          model: getConfigKaiProviderModel(),
+          parameters: getConfigKaiProviderParameters(),
```
ibolton336, line 141: Something seems to fail when passing in the parameters argument for me:

```ERROR - 2024-12-03 13:50:01,363 - kai.kai.reactive_codeplanner.task_manager.task_manager - Thread-1 - [task_manager.py:112 - supply_result()] - Encountered errors: ["Completions.create() got an unexpected keyword argument 'parameters'"]```

Comment (sjd78) [reply vscode/src/client/analyzerClient.ts:141]: Ian which kai rpc binary are you using?  This is something I can't test w/o linux binaries.

File: vscode/src/client/analyzerClient.ts (hunk 3 above)
Comment (sjd78) [reply vscode/src/client/analyzerClient.ts:433]: @djzager - The `initilizeParams.analyzer_lsp_rules_path` where this function is being called...can that be an array of strings on the `initialize` request or does the array need to be joined before the request?

Comment (djzager) [reply vscode/src/client/analyzerClient.ts:433]: Looks like I had gotten confused. The analyzer itself can take multiple paths, but the kai-analyzer-rpc (and kai-rpc-server by extension) only takes a single path.

I'll fix this now, once I have a tracking issue in kai for the future.

File: vscode/src/client/analyzerClient.ts (hunk 4 above)
Comment (ibolton336) [reply vscode/src/client/analyzerClient.ts]: Yeah that is the one @sjd78. Using an m1 mac luckily.

File: vscode/package.json [hunk 5]
```diff
@@ -403,33 +410,62 @@
...
+          "description": "URL for the Kai solution server",
+          "scope": "window"
+        },
+        "konveyor.kai.providerArgs": {
+          "type": "object",
+          "default": {
+            "model_id": "meta-llama/llama-3-1-70b-instruct",
+            "parameters": {
+              "max_new_tokens": 2048
+            }
```
ibolton336, line 465: During my tests I need this to be:
```suggestion
            "model": "gpt-3.5-turbo"
       
```
Else I get errors like:

```Encountered errors: ["Completions.create() got an unexpected keyword argument 'model-id'"]```

ibolton336 (approved):

vscode/package.json
sjd78: Hopefully these model configs don’t become a moving target. I do get the feeling it’ll end up a generic JSON “clob” to handle variations between models on the far back end.

PR #127, opened 2024-12-04
:sparkles: Add server status component

ibolton336:
- Disable analysis and getSolution if server is not running
- Move server status and prevent blocking analysis sidebar when server is starting
- Fix webview ready message to follow "type" convention
- Change startAnalysis to runAnalysis across all files

https://github.com/user-attachments/assets/d68bb713-8fde-482a-94fa-f4cbc3860086

### DIFF
--- a/vscode/src/KonveyorGUIWebviewViewProvider.ts
+++ b/vscode/src/KonveyorGUIWebviewViewProvider.ts
@@ -213,7 +213,7 @@ export class KonveyorGUIWebviewViewProvider implements WebviewViewProvider {
 
     webview.onDidReceiveMessage(
       (message) => {
-        if (message.command === "WEBVIEW_READY") {
+        if (message.type === "WEBVIEW_READY") {
           this._isWebviewReady = true;
           this._isPanelReady = true;
           while (this._messageQueue.length > 0) {
--- a/vscode/src/KonveyorGUIWebviewViewProvider.ts
+++ b/vscode/src/KonveyorGUIWebviewViewProvider.ts
@@ -136,7 +136,7 @@ export class KonveyorGUIWebviewViewProvider implements WebviewViewProvider {
         ${this._getReactRefreshScript(nonce)}
         <script nonce="${nonce}">
           window.addEventListener('DOMContentLoaded', function() {
-            window.vscode.postMessage({ command: 'webviewReady' });
+            window.vscode.postMessage({ type: 'WEBVIEW_READY' });
           });
         </script>
         <script type="module" nonce="${nonce}" src="${scriptUri}"></script>
@@ -213,7 +213,7 @@ export class KonveyorGUIWebviewViewProvider implements WebviewViewProvider {
 
     webview.onDidReceiveMessage(
       (message) => {
-        if (message.command === "webviewReady") {
+        if (message.command === "WEBVIEW_READY") {
           this._isWebviewReady = true;
           this._isPanelReady = true;
           while (this._messageQueue.length > 0) {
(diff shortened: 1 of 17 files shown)


vscode/src/client/analyzerClient.ts
```diff
@@ -58,13 +58,14 @@ export class AnalyzerClient {
...
   public async start(): Promise<void> {
+    console.log("AnalyzerClient.start()");
     if (!this.canAnalyze()) {
       vscode.window.showErrorMessage("Cannot start the server due to missing configuration.");
       return;
     }
 
     try {
-      Promise.all([exec("java -version"), exec("mvn -version")]);
+      await Promise.all([exec("java -version"), exec("mvn -version")]);
```
rszwajko, line 68: good catch!

vscode/src/KonveyorGUIWebviewViewProvider.ts
```diff
@@ -213,7 +213,7 @@ export class KonveyorGUIWebviewViewProvider implements WebviewViewProvider {
 
     webview.onDidReceiveMessage(
       (message) => {
-        if (message.command === "webviewReady") {
+        if (message.command === "WEBVIEW_READY") {
```
rszwajko, line 216: 1. `message .type` ? 
2. webview seemed to work without this notification - do we still need it?

vscode/src/client/analyzerClient.ts
```diff
@@ -276,11 +281,11 @@ export class AnalyzerClient {
         },
       );
 
-      console.log("response", response, incident, state);
-      vscode.commands.executeCommand("konveyor.loadSolution", response, { incident, violation });
+      vscode.commands.executeCommand("konveyor.loadSolution", response, {
```
rszwajko, line 284: it seems we have different formatting settings in the IDE

rszwajko (commented):
Looks good !

vscode/src/KonveyorGUIWebviewViewProvider.ts
ibolton336: The webview_ready is used by the panels - need to look at this when fixing the fullscreen analysis page.

vscode/src/client/analyzerClient.ts
ibolton336: Using prettier on my end. Not sure if there is a default prettier config that comes with the prettier plugin that may differ.

rszwajko: the line length is configured in the `.editorconfig` 
https://github.com/konveyor/editor-extensions/blob/9e3c6f3196a9fbb703124c57d403cd5ff15bd5bf/.editorconfig#L13

rszwajko:
@ibolton336  
I wonder if we really need to surface the server state - we could keep it internal and start it automatically if user requests analysis to be run.
Looking at the wireframes we might even drop the button (I understand it would then be run automatically )

sjd78:
I like having explicit status for the server available.  Back on the wireframes, I thought of it more as a status bar item with [wireframe 10](https://github.com/konveyor/editor-extensions/blob/main/docs/roadmap/wireframes/README.md#wf-10) (described but not drawn -- it is the blue bar in the status area in the wireframe image below).

[Wireframe 14](https://github.com/konveyor/editor-extensions/blob/main/docs/roadmap/wireframes/README.md#wf-14) has the progression from not configured forward...
image

In terms of UX, the server should manage itself like any other language server would.  If things are configured when the extension initializes, just start the server up right away.  The user changes the configs? Restart the server.  The extensions gets reloaded? Stop and start the server again.

For our purposes, manually starting and stopping the server can make a lot of sense as the start and initalize code can be changing.  Keeping those actions around and having a full set of server actions start, restart, and stop that can be hit from the command pallet is very vscode UX consistent.  Showing server status in the status bar is also far more vscode UX consistent.

djzager (approved):
ACK. I think we can iterate on this server status as we getting working on more systems more reliably in the coming weeks.

Issue #135, opened 2024-12-05
:bug: webview - server status easily broken


djzager:
The server status bar on the issues webview can get out-of-sync with the actual state of the server.

The `analyzerClient` needs to do a better job of tracking the state of the kai rpc server process.  Something like:
  - initial
  - configuration needed
  - ~ready to start~ configuration ready
  - starting
  - ~initializing~ ready to initialize
  - start failed
  - running
  - stopping
  - stopped

The webviews should then also track these states properly.

PR #138, opened 2024-12-06
:sparkles: Add commands for server stop and restart

sjd78:
- New commands to be able to stop or restart (stop then start) the kai rpc server are now available via the command pallet

- The server start now waits for the process to be spawned (or in error) before continuing to setup the server

- Once running the server start will wait for either 5 seconds, or for the server to report it is ready. After that race, the server is considered started and ready to be initialized.

- Server status is changed to "stopped" when the spawned server process is fully closed.

### DIFF
--- a/vscode/src/client/analyzerClient.ts
+++ b/vscode/src/client/analyzerClient.ts
@@ -83,19 +83,25 @@ export class AnalyzerClient {
     this.outputChannel.appendLine(`server args:`);
     this.getKaiRpcServerArgs().forEach((arg) => this.outputChannel.appendLine(`   ${arg}`));
 
-    this.kaiRpcServer = spawn(this.getKaiRpcServerPath(), this.getKaiRpcServerArgs(), {
+    const kaiRpcServer = spawn(this.getKaiRpcServerPath(), this.getKaiRpcServerArgs(), {
       cwd: serverCwd,
       env: this.getKaiRpcServerEnv(),
     });
+    this.kaiRpcServer = kaiRpcServer;
 
-    this.kaiRpcServer.on("spawn", () => {
-      this.outputChannel.appendLine(`kai rpc server has been spawned! [${this.kaiRpcServer?.pid}]`);
-    });
+    const pid = await new Promise<number | undefined>((resolve, reject) => {
+      kaiRpcServer.on("spawn", () => {
+        this.outputChannel.appendLine(
+          `kai rpc server has been spawned! [${this.kaiRpcServer?.pid}]`,
+        );
+        resolve(this.kaiRpcServer?.pid);
+      });
 
-    this.kaiRpcServer.on("error", (err) => {
-      this.outputChannel.appendLine(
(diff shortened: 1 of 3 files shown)


vscode/src/client/analyzerClient.ts
```diff
@@ -117,19 +131,28 @@ export class AnalyzerClient {
...
+      new Promise<void>((resolve) => {
+        kaiRpcServer.on("serverReportsReady", (pid) => {
+          this.outputChannel.appendLine(`*** kai rpc server [${pid}] reports ready`);
+          resolve();
+        });
+      }),
+      setTimeout(5000),
+    ]);
+
+    this.outputChannel.appendLine(`Started the kai rpc server, pid: [${pid}]`);
```
djzager, line 144: Clever

vscode/src/commands.ts
```diff
@@ -51,12 +51,37 @@ const commandsMap: (state: ExtensionState) => {
...
-
       try {
         await analyzerClient.start();
         await analyzerClient.initialize();
       } catch (e) {
-        console.error("Could not start the analyzer", e);
+        console.error("Could not start the server", e);
+      }
+    },
+    "konveyor.stopServer": async () => {
```
djzager, line 61: Do we want to tie this into the stop server button in the webview or create a GH issue to track that?

djzager (approved):
LGTM.

vscode/src/commands.ts
sjd78: I figure any updates to the web view would be better in another pr

vscode/src/client/analyzerClient.ts
```diff
@@ -83,19 +83,25 @@ export class AnalyzerClient {
...
-    });
-
-    this.kaiRpcServer.on("error", (err) => {
-      this.outputChannel.appendLine(
-        `[error] - error in process[${this.kaiRpcServer?.spawnfile}]: ${err}`,
-      );
+      kaiRpcServer.on("error", (err) => {
+        const message = `error in process[${this.kaiRpcServer?.spawnfile}]: ${err}`;
+        this.outputChannel.appendLine(`[error] - ${message}}`);
+        reject();
```
rszwajko, line 103: `reject`  will break the flow - we should probably add `catch` block that would i.e. set the server status (I believe this kind of problems motivated https://github.com/konveyor/editor-extensions/issues/135)

vscode/src/client/analyzerClient.ts
sjd78: Old pattern that I've used before to cancel an operation that's running for too long.  Different use here, but still works well.

vscode/src/client/analyzerClient.ts
sjd78: That is part of it for sure.  But really the server's `ServerState` should be tracked much more explicitly than it is now.  There are some gaps in potential.  I can address those in another followup PR.

PR #139, opened 2024-12-06
:bug: make analysis runs cancellable

### LINKED ISSUE #131: User should be able to cancel an analysis
Currently, when the user chooses to run analysis the "process" is not cancellable. This makes it so that when something goes wrong, the only recourse is for the user to restart vscode extension host.

djzager:
Fixes #131

### DIFF
--- a/vscode/src/client/analyzerClient.ts
+++ b/vscode/src/client/analyzerClient.ts
@@ -236,9 +236,9 @@ export class AnalyzerClient {
       {
         location: vscode.ProgressLocation.Notification,
         title: "Running Analysis",
-        cancellable: false,
+        cancellable: true,
       },
-      async (progress) => {
+      async (progress, token) => {
         try {
           progress.report({ message: "Running..." });
           this.fireAnalysisStateChange(true);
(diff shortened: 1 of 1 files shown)


vscode/src/client/analyzerClient.ts
```diff
@@ -253,10 +253,16 @@ export class AnalyzerClient {
             )}`,
           );
 
-          const response: any = await this.rpcConnection!.sendRequest(
-            "analysis_engine.Analyze",
-            requestParams,
-          );
+          const cancellationPromise = new Promise((_, reject) => {
+            token.onCancellationRequested(() => {
+              reject(new Error("Operation canceled by the user."));
```
rszwajko, line 258: 1. rejecting is equivalent to throwing an exception -  I cannot test it but the `catch` block should get called. Consider resolving and exiting in a controlled way.
2. in the vscode code base  `token.isCancellationRequested` check is used sometimes to support the callback (I guess just-in-case the action would be cancelled before the callback was added)

djzager: Good catch. Addressed both.

vscode/src/client/analyzerClient.ts
```diff
@@ -279,10 +279,22 @@ export class AnalyzerClient {
...
+            return;
+          }
+
+          const cancellationPromise = new Promise((_, reject) => {
+            token.onCancellationRequested(() => {
+              reject(new Error("Operation canceled by the user."));
+            });
+          });
+
+          const response: any = await Promise.race([
```
rszwajko, line 294: I'd rather resolve the promise i.e.:
```
const {response, isCancelled } = await Promise.race(....
```
and branch based on the returned flag.

rszwajko: @djzager 
I've added a commit with the fix to this PR

ibolton336 (approved):

sjd78 (approved):
So technically the analysis process won't actually get cancelled, but the UI will be unblocked.

Good enough for now.

PR #143, opened 2024-12-09
:ghost: Add configuration command and option for GenAIKey

ibolton336:
https://github.com/user-attachments/assets/696e7822-fae0-4d80-912c-4045c39a7720


vscode/src/utilities/configuration.ts
```diff
@@ -144,3 +144,11 @@ export async function updateKaiProviderName(value: string): Promise<void> {
 export async function updateKaiProviderModel(value: string): Promise<void> {
   await updateConfigValue("kai.providerModel", value, vscode.ConfigurationTarget.Workspace);
 }
+
+export function getConfigKaiGenAiKey(): string {
+  return vscode.workspace.getConfiguration(KONVEYOR_CONFIG_KEY)?.get<string>("kai.genAiKey") || "";
+}
+export async function updateGenAiKey(newKey: string | undefined): Promise<void> {
+  const config = vscode.workspace.getConfiguration(KONVEYOR_CONFIG_KEY);
+  await config.update("kai.genAiKey", newKey, vscode.ConfigurationTarget.Global);
```
rszwajko, line 153: we should keep the key encrypted - as temporary solution we can use [`context.secrets.store("key", value)`](https://code.visualstudio.com/api/references/vscode-api#SecretStorage)

rszwajko (changes requested):

rszwajko (approved):

PR #147, opened 2024-12-10
:sparkles: Add configuration for getSolution params

ibolton336:
https://github.com/user-attachments/assets/b8932454-7255-4c32-88b5-f02a2c581150

### DIFF
--- a/vscode/media/walkthroughs/configure-get-Solution-params.md
+++ b/vscode/media/walkthroughs/configure-get-Solution-params.md
@@ -1,0 +1,3 @@
+# Configure Get Solution Params
+
+The Konveyor extension allows users to configure the parameters for the Get Solution command. Max_depth, max_priority, and max_iterations are the parameters that can be configured. The Get Solution command is used to get the solution for the given project based on the analysis results.
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -220,6 +220,12 @@
         "icon": "$(arrow-left)",
         "category": "diffEditor",
         "title": "Apply Selection"
+      },
+      {
+        "command": "konveyor.configureGetSolutionParams",
+        "title": "Configure getSolution Parameters",
+        "category": "Konveyor",
+        "icon": "$(gear)"
       }
     ],
     "submenus": [
@@ -504,12 +510,7 @@
         },
         "konveyor.kai.providerArgs": {
           "type": "object",
-          "default": {
-            "model_id": "meta-llama/llama-3-1-70b-instruct",
-            "parameters": {
-              "max_new_tokens": 2048
-            }
-          },
+          "default": {},
           "description": "Kai provider arguments",
           "scope": "window"
         },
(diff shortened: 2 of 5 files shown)


vscode/media/walkthroughs/configure-get-Solution-params.md
sjd78, line 1: This description will need to be revisited.

vscode/package.json
```diff
@@ -478,6 +484,21 @@
           },
           "description": "Kai provider arguments",
           "scope": "window"
+        },
+        "konveyor.kai.getSolutionMaxPriority": {
+          "type": "number",
+          "default": 0,
+          "description": "Maximum number of properties to get from Kai"
```
sjd78, line 491: Is this description correct?  Priority != properties

vscode/src/commands.ts
```diff
@@ -389,6 +392,57 @@ const commandsMap: (state: ExtensionState) => {
     "konveyor.diffView.applySelection": applyBlock,
     "konveyor.diffView.applySelectionInline": applyBlock,
     "konveyor.partialAnalysis": async (filePaths: string[]) => runPartialAnalysis(state, filePaths),
+    "konveyor.configureGetSolutionParams": async () => {
```
sjd78, line 395: Just a comment ---

I'm not a fan of the input boxes like this.  I'd rather have a configuration item link to settings.json directly and let code assist handle this kind of base level value.

But that does seem like a future thing to overhaul how the configurations work...

sjd78 (changes requested):
The code flow looks ok to me.

The descriptions will need rewording, or the entire way we handle configuring Kai settings will (hopefully eventually) be redone so it's not just a series of quick picks and input values.

Only blocking comment is the description for `getSolutionMaxPriority`.  If this is the correct description, comment as such and I'll be an approve.

vscode/src/client/analyzerClient.ts
```diff
@@ -324,19 +327,28 @@ export class AnalyzerClient {
...
 
+    const maxPriority = getConfigMaxPriority();
+    const maxDepth = getConfigMaxDepth();
+    const maxIterations = getConfigMaxIterations();
+
     try {
+      this.outputChannel.appendLine(`Sending 'getCodeplanAgentSolution' with:
+        max_priority: ${maxPriority}, 
+        max_depth: ${maxDepth}, 
+        max_iterations: ${maxIterations}`);
```
sjd78, line 342: Nit-pick: If the request payload is built outside the call, the full payload can be logged...

```ts
const request = {
  file_path: "",
  incidents: ...,
  max_priority: ...
  ...
};
this.outputChannel.appendLine(`getCodeplanAgentSolution request: ${JSON.stringify(request, null, 2)}`);

....sendRequest("getCodeplanAgentSolution", request);
```

sjd78 (commented):
WDYT about the request logging nit-pick?  Everything else looks great.

sjd78 (approved):
LGTM!

PR #151, opened 2024-12-11
:sparkles: Remove default toml arguments and use settings.json

ibolton336:
- Toml values overwrite configured settings. Removing these values for now. 
- Add in defaults for providerArgs
- Allow demoMode to be set from settings.json

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -510,7 +510,12 @@
         },
         "konveyor.kai.providerArgs": {
           "type": "object",
-          "default": {},
+          "default": {
+            "model_id": "meta-llama/llama-3-1-70b-instruct",
+            "parameters": {
+              "max_new_tokens": 2048
+            }
+          },
           "description": "Kai provider arguments",
           "scope": "window"
         },
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -525,6 +525,12 @@
           "description": "Generative AI Key",
           "scope": "window"
         },
+        "konveyor.kai.demoMode": {
+          "type": "boolean",
+          "default": false,
+          "description": "Demo mode for Kai",
+          "scope": "window"
+        },
         "konveyor.kai.getSolutionMaxPriority": {
           "type": "number",
           "default": 0,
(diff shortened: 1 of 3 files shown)


vscode/package.json
```diff
@@ -520,6 +525,12 @@
           "description": "Generative AI Key",
           "scope": "window"
         },
+        "konveyor.kai.demoMode": {
+          "type": "boolean",
+          "default": false,
+          "description": "Demo mode for Kai",
+          "scope": "window"
+        },
```
sjd78, line 533: Setting "konveyor.kai.demoMode" in settings.json did work, it just wasn't listed so it wasn't a documented config key.

Here is the previous discussion: https://github.com/konveyor/editor-extensions/pull/126#discussion_r1867961970

sjd78 (commented):
The toml / default model config values changes LGTM.

Do we really want the demoMode setting to be visible in the settings panel?

vscode/package.json
```diff
@@ -510,7 +510,12 @@
         },
         "konveyor.kai.providerArgs": {
           "type": "object",
-          "default": {},
+          "default": {
```
ibolton336, line 513: Adding back in for run_demo support. TODO - need to allow default **and** overwriting / merging of all of these arguments when changing providers.

vscode/package.json
ibolton336: Closing RE: solution discovered on slack.

sjd78 (approved):
LGTM

vscode/package.json
sjd78: Yeah I get the feeling that the standard vscode config support isn't going to be enough to handle the various permutations we want to be able to cover.  There is capability to be able to setup a custom editor (a webview really) for specific files.  So maybe in the near future, the kai server and model configs will be in a "custom" config file with a custom editor...

ibolton336: @sjd78 https://github.com/konveyor/editor-extensions/pull/153 - something to get us by for now.

PR #153, opened 2024-12-11
:bug: Replace all default args if any provider args set

ibolton336:
- Prevent default args passed to incompatible models when user is setting a custom provider

### DIFF
--- a/vscode/src/client/analyzerClient.ts
+++ b/vscode/src/client/analyzerClient.ts
@@ -177,6 +177,14 @@ export class AnalyzerClient {
       return;
     }
 
+    // Get the provider args from the user settings or the default settings
+    const config = vscode.workspace.getConfiguration("konveyor.kai");
+    const userProviderArgs = getConfigKaiProviderArgs();
+    const finalProviderArgs = userProviderArgs || config.get<object>("providerArgs");
+    this.outputChannel.appendLine(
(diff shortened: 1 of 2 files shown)


vscode/src/client/analyzerClient.ts
```diff
@@ -177,6 +177,14 @@ export class AnalyzerClient {
       return;
     }
 
+    // Get the provider args from the user settings or the default settings
+    const config = vscode.workspace.getConfiguration("konveyor.kai");
+    const userProviderArgs = getConfigKaiProviderArgs();
+    const finalProviderArgs = userProviderArgs || config.get<object>("providerArgs");
+    this.outputChannel.appendLine(
+      `Using finalProviderArgs: ${JSON.stringify(finalProviderArgs, null, 2)}`,
+    );
```
sjd78, line 186: This probably would look a bit better if it followed the `isDemoMode()` protected function style and build the entire `model_provider` sectin.  That way we can have a single function whose job it is to build the `model_provider`

sjd78 (commented):
The config function looks ok.

Having a `buildModelProvider()` function in the same spirit as `isDemoMode()` would make me happy.

vscode/src/client/analyzerClient.ts
sjd78: Just need to remove this block now that it is in the function

sjd78 (approved):
LGTM

PR #157, opened 2024-12-12
:seedling: Use kai binaries from release v0.0.3

sjd78:
- Use release v0.0.3

- Adjust `collect-assets.js` to use the "new" asset names

- Use case insensitive asset name matching between release
  assets and the configurations

- Add linux and windows arm64 builds


savitharaghunathan (approved):
visiack

ibolton336 (approved):

PR #161, opened 2024-12-12
:bug: Escape initial state data injected into webview

rszwajko:


rszwajko:
### Example failure
Incident data contained tags that were rendered instead of the webview
Screenshot from 2024-12-12 21-01-33

ibolton336 (approved):

PR #162, opened 2024-12-13
:bug: Ignore results outside of requested analysis scope

rszwajko:
Additional duplicated results were accumulated in the UI.


djzager (approved):
lgtm

Issue #133, opened 2024-12-05
:bug: Extension should handle rpc process(es) being killed


djzager:
If a `killall kai-rpc-server` is run, the extension should end up knowing the server is "stopped".

PR #166, opened 2024-12-13
:sparkles: Enhance server process and status handling

sjd78:
Resolves: #133
Resolves: #135

Use additional child_process eventing to keep track of the server process itself.  This lets the extension time the initialize call better, and be able to "cleanly" shutdown the `kai-rpc-server``.  If the `kai-rpc-server` is killed or terminated externally, it will be noticed and the extension will understand the server is stopped.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "editor-extensions",
-  "version": "0.0.1",
+  "version": "0.0.3",
   "lockfileVersion": 3,
   "requires": true,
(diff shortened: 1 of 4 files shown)


djzager (approved):
lgtm

image

sjd78:
> lgtm
> image

You'll probably want to also keep an eye out for rogue kai-analyzer-rpc and java (jdtls) processes.  Killing the kai-rpc-server doesn't seem to kill the child processes.

Issue #160, opened 2024-12-12
v0.0.3 - In a clean workspace (no existing .vscode/konveyor folder), starting the server fails


sjd78:
Release: v0.0.3 / main
Runtime method: unzipped vsix in the .vscode/extensions folder of my workspace

Server fails to start with output channel logging:
```
Starting the kai rpc server ...
server cwd: /home/sdickers/projects/__kai__coolstore/.vscode/konveyor/kai-runtime
server path: /home/sdickers/projects/__kai__coolstore/.vscode/extensions/extension/assets/kai/linux-x64/kai-rpc-server
server args:
   --config
   /home/sdickers/projects/__kai__coolstore/.vscode/konveyor/kai/kai-config.toml
kai rpc server has been spawned! [1095116]
pydantic/_internal/_fields.py:161: UserWarning: Field "model_provider" has conflict with protected namespace "model_".

You may be able to resolve this warning by setting `model_config['protected_namespaces'] = ()`.

Traceback (most recent call last):
  File "kai/rpc_server/main.py", line 85, in <module>
  File "kai/rpc_server/main.py", line 53, in main
  File "kai/kai_config.py", line 289, in model_validate_filepath
  File "pydantic_settings/main.py", line 144, in __init__
  File "pydantic/main.py", line 193, in __init__
pydantic_core._pydantic_core.ValidationError: 1 validation error for KaiConfig
models
  Field required [type=missing, input_value={'log_level': 'info', 'fi...e/.vscode/konveyor/kai'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.8/v/missing
[PYI-1095117:ERROR] Failed to execute script 'main' due to unhandled exception!

kai rpc server exited with signal null, code 1
kai rpc server closed with signal null, code 1
Started the kai rpc server, pid: [1095116]
initialize payload: {
  "process_id": null,
  "kai_backend_url": "0.0.0.0:8080",
  "root_path": "/home/sdickers/projects/__kai__coolstore",
  "log_level": "debug",
  "log_dir_path": "/home/sdickers/projects/__kai__coolstore/.vscode/konveyor/kai",
  "model_provider": {
    "provider": "ChatIBMGenAI",
    "args": {
      "model_id": "meta-llama/llama-3-1-70b-instruct",
      "parameters": {
        "max_new_tokens": 2048
      }
    }
  },
  "file_log_level": "debug",
  "demo_mode": false,
  "cache_dir": "",
[...]

PR #168, opened 2024-12-13
:bug: Add back default config.toml values

sjd78:
Resolves: #160

Pending https://github.com/konveyor/kai/pull/526, this change is needed so the kai-rpc-server starts.  The initialize call should still override any model configurations in the toml file.

Once the kai change to remove `config.toml` is merged, the TODOs added by this change will need to be resolved.


ibolton336:
You can just add the values into the settings.json for providerArgs and provider name to fix this. No need to add to the config.toml defaults.

djzager (approved):

djzager:
Confirmed this is an issue.

Issue #171, opened 2025-01-06
[Feature] Generate fix for Violation (all incidents of a particular violation)


djzager:
# Description:

Implement a "Fix All Incidents" feature to allow users to generate a fix for a violation and all of its associated incidents. This will involve:

- Adding a button to the webview.
- Adding a button to the Konveyor Issues Panel.
- (Optionally) Adding a way to invoke this feature from the native VS Code Problems Panel.
- Collecting all incidents for the violation and passing them to getSolution.
- Modifying maxPriority, maxDepth, and maxIterations before making the request to Kai for fix generation.

# Acceptance Criteria:

- Buttons in relevant panels invoke the "Fix All Incidents" functionality.
- All incidents of the selected violation are included in the request to getSolution.
- Parameters (maxPriority, maxDepth, maxIterations) are correctly updated before the fix request.
- The fix is generated and applied across all incidents seamlessly.

# Change needs
  - [ ] Analysis view ([wireframe 17](https://github.com/konveyor/editor-extensions/blob/main/docs/roadmap/wireframes/17.issues-view-webview.excalidraw.svg) has design details)
    - [ ] needs action icons at different scope levels
    - [ ] needs tooltips on the action icons to indicate how many incidents will be included in the solution request (Note: The UI views say "resolution" while the api call is "solution". The UI should be consistent and use "resolution" only in text labels and messages.)
      - "Resolve this incident"
      - "Resolve the 20 incidents within this issue"
    - [ ] needs, on a hover over a view action, a visual indicator on the issue or incident grouping to indicate the scope of the resolution

  - [x] Solution request command and api call need to support an array of incidents in place of a single incident (see #186)

  - [ ] Resolution view ([wireframe 19](https://github.com/konveyor/editor-extensions/blob/main/docs/roadmap/wireframes/19.resolution-view.excalidraw.svg))
    - [ ] Needs to support showing the solution request that contains multiple incidents

# In Future
The Konveyor Issues tree view can be enhanced to add a request resolution on the various levels of the tree matching the levels in the webview.

djzager:
Priority is on the webview.

PR #186, opened 2025-01-09
:seedling: Refactor getSolution command to accept a list of incidents

rszwajko:
Reference-Url: https://github.com/konveyor/editor-extensions/issues/171

### DIFF
--- a/webview-ui/src/components/ResolutionsPage.tsx
+++ b/webview-ui/src/components/ResolutionsPage.tsx
@@ -131,8 +131,8 @@ const ResolutionPage: React.FC = () => {
               </CardTitle>
               <CardBody>
                 <IncidentList
-                  incidents={solutionScope?.incident ? [solutionScope?.incident] : []}
-                  selectedIncident={solutionScope?.incident}
+                  incidents={solutionScope?.incidents ?? []}
+                  selectedIncident={solutionScope?.incidents?.[0]}
                   onSelectIncident={() => {}}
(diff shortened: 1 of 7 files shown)


djzager (commented):
lgtm

webview-ui/src/components/ResolutionsPage.tsx
```diff
@@ -131,8 +131,8 @@ const ResolutionPage: React.FC = () => {
               </CardTitle>
               <CardBody>
                 <IncidentList
-                  incidents={solutionScope?.incident ? [solutionScope?.incident] : []}
-                  selectedIncident={solutionScope?.incident}
+                  incidents={solutionScope?.incidents ?? []}
+                  selectedIncident={solutionScope?.incidents?.[0]}
```
sjd78, line 135: Looks like this will need to updated once multiple incidents are actually requested.  The resolution view needs to be able to handling showing a set of incidents that have been requested to be fixed.

Probably need an issue to track this.

sjd78 (approved):
The base changes look good.

It'll need followup work in the resolution webview to handle displaying multiple incident solution requests.

webview-ui/src/components/ResolutionsPage.tsx
rszwajko: It seems that without this change we won't be able to view the resolutions so I will fix it as part of the #171

PR #188, opened 2025-01-09
:seedling: update vscode extension readme

djzager:

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -1,5 +1,10 @@
 {
-  "name": "editor-extensions-vscode",
+  "name": "konveyor-ai",
+  "displayName": "Konveyor AI (kai) Extension for VSCode",
+  "description": "Generative AI assisted migration and modernization tool",
+  "author": "Konveyor",
+  "icon": "resources/konveyor-icon-color.png",
+  "license": "Apache-2.0",
   "version": "0.0.4",
   "main": "./out/extension.js",
   "publisher": "konveyor",
@@ -8,6 +13,7 @@
     "url": "https://github.com/konveyor/editor-extensions"
   },
   "bugs": "https://github.com/konveyor/editor-extensions/issues",
+  "preview": false,
   "categories": [
     "Programming Languages",
     "Machine Learning",
--- a/package-lock.json
+++ b/package-lock.json
@@ -4718,10 +4718,6 @@
         "safe-buffer": "^5.0.1"
       }
     },
-    "node_modules/editor-extensions-vscode": {
-      "resolved": "vscode",
-      "link": true
-    },
     "node_modules/electron-to-chromium": {
       "version": "1.5.62",
       "resolved": "https://registry.npmjs.org/electron-to-chromium/-/electron-to-chromium-1.5.62.tgz",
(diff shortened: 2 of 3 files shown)


dymurray (approved):

vscode/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "editor-extensions-vscode",
+  "name": "kai-vscode",
```
sjd78, line 2: Do we want this name to match the target rename in the release workflow and thn just drop the rename?

https://github.com/konveyor/editor-extensions/blob/74a357ef8e67c5c3537a10669c42c040d50ada6a/.github/workflows/release.yml#L141-L154

sjd78: Also, there are a few attributes that we should add that I was looking at in the java extension's package.json: 

https://github.com/redhat-developer/vscode-java/blob/fdfccb29a6dbafaf4cda35ff7693e05bfd0f4eeb/package.json#L1-L11
```json
{
  "name": "java",
  "displayName": "Language Support for Java(TM) by Red Hat",
  "description": "Java Linting, Intellisense, formatting, refactoring, Maven/Gradle support and more...",
  "author": "Red Hat",
  "icon": "icons/icon128.png",
  "license": "EPL-2.0",
  "version": "1.39.0",
  "publisher": "redhat",
  "bugs": "https://github.com/redhat-developer/vscode-java/issues",
  "preview": false,
  "capabilities": {
    "untrustedWorkspaces": {
      "supported": "limited",
      "restrictedConfigurations": [
        "java.jdt.ls.java.home",
        "java.home",
        "java.jdt.ls.vmargs",
        "java.configuration.runtimes",
        "java.import.gradle.java.home",
        "java.import.gradle.home"
      ]
    },
    "virtualWorkspaces": false
  },
  "<snip>": "<snip>"
}
```

sjd78 (changes requested):
The readme looks pretty good.

It would be nice, in future, to include mini feature demos that match the current state of the extension.  This is fairly standard practice for vscode and extensions.  Plenty of examples looking at the extensions view in vscode.

Requesting changes just for the one title in the readme and to discuss the name, displayName, release file name stuff.

vscode/package.json
djzager: @dymurray, thoughts on this? It sounded like you wanted `kai` in there.

djzager: Do we have need for the `.capabilities`? I am slow and I don't see what of those we should define.

sjd78: Maybe a name of "konveyor-ai" and a displayName of "Konveyor AI (Kai) Support"?

Linking the icon we already have would be nice (assuming it would work).

sjd78: > Do we have need for the `.capabilities`? I am slow and I don't see what of those we should define.

I think we’re ok without it for now. Of need to read up on what can be in there. Should be easy enough to add when needed. Feels like something that comes before maybe a beta release.

djzager:
image

vscode/package.json
djzager: I think I got it all. Let me know if I missed anything.

sjd78 (approved):
Looks good to me.  The new landing page and the extra meta-info in package.json look nice.

PR #195, opened 2025-01-13
:sparkles: Get solutions for violation and a group of incidents

rszwajko:
Part of https://github.com/konveyor/editor-extensions/issues/171


rszwajko:
## Issue web view

1. `Fix violation` - top level
2. `Fix group of incidents`
3. `Fix incident` - lowest level (table actions)

Screenshot from 2025-01-14 12-09-01


https://github.com/user-attachments/assets/36d835ab-a3c0-4606-93d0-8e8e02f3a0ad



## Issue tree view

1. `Fix group of incidents`  - context action at the top level
4. `Fix incident` - context action at the lowest level - tooltip visible on the screenshot below

Screenshot from 2025-01-14 12-12-05

webview-ui/src/components/ViolationIncidentsList.tsx
```diff
@@ -211,14 +150,18 @@ const ViolationIncidentsList: React.FC<ViolationIncidentsListProps> = ({
...
-            // }}
+            actions={{
+              actions: (
+                <ViolationActionsDropdown
+                  onGetAllSolutions={() => onGetSolution(violation.incidents ?? [], violation)}
+                  fixMessage={
+                    violation.incidents?.length === 1
+                      ? "Resolve 1 incident within this issue"
+                      : `Resolve the ${violation.incidents?.length ?? 0} incidents within this issue`
+                  }
```
ibolton336, line 161: IMO "Within this issue"  may be borderline redundant & could do with a shortened version like

```suggestion
  fixMessage={`Resolve ${violation.incidents?.length === 1 ? '1 incident' : `${violation.incidents?.length ?? 0} incidents`}`}
```

webview-ui/src/components/ViolationIncidentsList.tsx
```diff
@@ -242,17 +185,22 @@ const ViolationIncidentsList: React.FC<ViolationIncidentsListProps> = ({
           </CardHeader>
           <CardExpandableContent>
             <CardBody>
-              {violation.incidents.map((incident) => (
-                <div key={`${incident.uri}-${incident.lineNumber}`}>
-                  {renderIncident(incident, violation)}
-                </div>
+              {Object.entries(groupIncidentsByMsg(violation.incidents)).map(([message, tuples]) => (
```
ibolton336, line 188: Would prefer to see the transformation done outside of the render method / JSX.

webview-ui/src/components/IncidentTable/IncidentTable.tsx
```diff
@@ -0,0 +1,101 @@
...
+                  incidents.length === 1
+                    ? "Resolve 1 incident"
+                    : `Resolve the ${incidents.length} incidents`
+                }
+              />
+            ),
+          }}
+        />
+
+        <CardBody>{message}</CardBody>
```
ibolton336, line 55: Could use a label for this message.

ibolton336 (changes requested):
See inline comments. Looking ok to me for the most part just a few qualms with style.

webview-ui/src/components/IncidentTable/IncidentTable.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const ISSUE = "Issue";
+  const LOCATION = "Location";
+  const FOLDER = "Folder";
+  return (
+    <>
+      <Card isPlain>
+        <CardHeader
+          actions={{
+            actions: (
+              <ViolationActionsDropdown
```
ibolton336, line 43: A style change may be warranted so that it is clear the table action scope is different than the outer violation action scope. Maybe a background color change for the table would do? @JustinXHale 
Screenshot 2025-01-14 at 3 55 15 PM

webview-ui/src/components/ViolationIncidentsList.tsx
rszwajko: 1. it's a feature :)  I agree the message is a bit verbose but it was requested this way [in the spec ](https://github.com/konveyor/editor-extensions/issues/171). @sjd78  are you OK with shortening the labels? Note that in tree view the labels are shorter because (by default) we cannot pass the number of items in the category (labels for context actions are defined in package.json)
2. in general concatenating strings breaks translation (we are English only now but it may change)

webview-ui/src/components/IncidentTable/IncidentTable.tsx
rszwajko: The label does have the "Show more" but also has a border that we would need to remove manually.  Using plain text seemed the closest to [wireframe 17](https://github.com/konveyor/editor-extensions/blob/main/docs/roadmap/wireframes/17.issues-view-webview.excalidraw.svg)

webview-ui/src/components/IncidentTable/IncidentTable.tsx
rszwajko: I think we are OK here. The most similar PF functionality would  be the [tree view with actions](https://www.patternfly.org/components/tree-view/#with-action-items). There the recommended approach is to place actions in one column. This was also the proposed layout in the [wireframe 17](https://github.com/konveyor/editor-extensions/blob/main/docs/roadmap/wireframes/17.issues-view-webview.excalidraw.svg) however  the current PR deviated a bit (see below for details).

In the wireframes:
1. issue/violation actions at the same level as group-of-incidents actions
2. only per-incident actions are at different position

In this PR the layout was  dictated by the components I've used:
1. group-of-incidents actions got moved (Card component default behavior)
2. table actions got pushed to right due to limited space - folder path takes a lot of space

webview-ui/src/components/ViolationIncidentsList.tsx
rszwajko: Addressed [in this commit](https://github.com/konveyor/editor-extensions/pull/195/commits/47235052d50a29532e29bad878e7bae43925f5a9).

webview-ui/src/components/IncidentTable/IncidentTable.tsx
JustinXHale: Will the Kebab have more than 1 action within it?

webview-ui/src/components/IncidentTable/IncidentTable.tsx
rszwajko: Implemented as suggested in commit https://github.com/konveyor/editor-extensions/pull/195/commits/97b41a69e9e14a1723db6a7dae6d9c1c08efe141

New look:
Screenshot from 2025-01-15 20-03-07

webview-ui/src/components/IncidentTable/IncidentTable.tsx
rszwajko: > Will there be more than one action on the kebab menu?
>
I assume we will have more actions there i.e. requesting resolution with different parameters ( config-based, custom/ad hoc, unlimited power). Currently there is one option.

> When the user clicks an action like "Resolve # items," would they be prompted for confirmation afterwards, or does it take effect immediately?
>
Action is executed immediately. The action is costly (LLM requests cost $) and may be long but we (will) provide a way to cancel it if necessary.

ibolton336 (approved):

webview-ui/src/components/ViolationIncidentsList.tsx
rszwajko: Resolving the comment to unblock the merge - we can improve the messages later on.

webview-ui/src/components/IncidentTable/IncidentTable.tsx
rszwajko: Thank you @JustinXHale  for your comments - the kebab seems justified, confirming the actions is sth that we need to discuss (especially when we still cannot cancel in-progress LLM processing).
CC: @sjd78 @djzager

Issue #184, opened 2025-01-08
[Tech Debt] Configuring AI Model + API Key is Painful


djzager:
Currently, configuring the AI model to use and saving the user's API Key is not pleasant. We should fix it.

dymurray:
Additionally; we don't currently have any way for a user to configure using openAI easily. The key for openAI is a different env var; not `GENAI_KEY`. The var needs to be `OPENAI_API_KEY`. We need to update the model configuration to at least handle the most common providers we intend to support.

sjd78:
Thoughts:
- runtime settings in `settings.json`...
  - settings for at least log level, binary locations, custom rule locations, analysis arguments, get solution knobs
  - uses standard vscode config definitions and UI (`Ctrl+,` or command `Preferences: Open Workspace Settings (settings.json)`
 
- still use the extension's walkthrought to point out the main config needs (and use command to update from there w/o going through the normal settings editor)
 
- model settings in an extension specific config file that lives outside of the repo
  - have a command that opens the file for editing
  - base editor can have a schema associated with it to provide traditional code assist help with configs (and/or include code snippets that template configs for different models)
  - in future, may add a webview for editing it to support a richer/guided UX

Look in to:
  - Can the settings config descriptions include a link to the commands that can help set the values?  This would allow the command use both from the settings UI and from the walkthroughs.
  - Can there be an info/placeholder config item with a link to open the extension's ai config file?  This would allow all the settings to be visible in the normal settings UI but push the user to our custom file.

sjd78:
> Do we need the ability to say "X config is here" meaning an environment variable or some other file?

I'm stuck in an ambiguity loop trying to understand your meaning here.  Anything that needs to be configured for the kai server and anything it uses needs to be captured in a config setting somewhere under the extension's purview.  If one of the configs is "use the file at this path" then ok, we'll capture that as a file path.  I'm not particularly worried about that.

I'm thinking the configs the extension stores will be passed along as either an [ENVVAR](https://github.com/konveyor/editor-extensions/blob/main/vscode/src/client/analyzerClient.ts#L606-L616), a [cli argument](https://github.com/konveyor/editor-extensions/blob/main/vscode/src/client/analyzerClient.ts#L631-L635), or in the [initialize request](https://github.com/konveyor/editor-extensions/blob/main/vscode/src/client/analyzerClient.ts#L251-L277).  We have well defined places in `analyzerClient.ts` where those are created, so updating them as needed should be straightforward.

[...]

PR #190, opened 2025-01-10
:ghost: Update analyzerClient.ts

JonahSussman:
Update initialize parameters to align with https://github.com/konveyor/kai/pull/526. Please feel free to commit to this PR.

PR #197, opened 2025-01-15
:seedling: Update config definitions, base configs and initialize call

sjd78:
Based on changes to the kai runtime configurations in https://github.com/konveyor/kai/pull/526, update the CLI arguments and initialize request object to match.

Summary of changes:
  - Reorder the config field definitions

  - Use enum values in config definitions where available

  - Config keys related to analyzer configuration still remain but cannot be passed along to the analyzer, see [kai issue 550](https://github.com/konveyor/kai/issues/550)

  - Aligned CLI arguments and initialize request payload types to the updated Kai configuration keys

  - Dropped config.toml handling since that file is no longer in use

Prerequisite for: #184

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -409,12 +409,27 @@
       "type": "object",
       "title": "Konveyor",
       "properties": {
+        "konveyor.logLevel": {
+          "type": "string",
+          "enum": [
+            "TRACE",
+            "DEBUG",
+            "INFO",
+            "WARN",
+            "ERROR",
+            "CRITICAL"
+          ],
+          "default": "DEBUG",
+          "description": "Log level to use with extension",
+          "scope": "window",
+          "order": 0
+        },
         "konveyor.analyzerPath": {
           "type": "string",
           "default": "",
           "description": "Path to the analyzer binary. If not set, the extension will use the bundled analyzer.",
           "scope": "machine",
-          "order": 0
+          "order": 2
         },
         "konveyor.kaiRpcServerPath": {
           "type": "string",
@@ -423,40 +438,12 @@
           "scope": "machine",
           "order": 1
         },
-        "konveyor.logLevel": {
-          "type": "string",
-          "default": "debug",
-          "description": "Log level to use with extension",
-          "scope": "window",
-          "order": 2
-        },
-        "konveyor.analysis.incidentLimit": {
-          "type": "number",
-          "default": 10000,
-          "description": "Maximum number of incidents to report",
-          "scope": "window",
-          "order": 1
-        },
-        "konveyor.analysis.contextLines": {
-          "type": "number",
-          "default": 10,
-          "description": "Number of lines of context to include in incident reports",
-          "scope": "window",
-          "order": 2
-        },
-        "konveyor.analysis.codeSnipLimit": {
-          "type": "number",
-          "default": 10,
-          "description": "Number of lines of code to include in incident reports",
-          "scope": "window",
-          "order": 3
-        },
         "konveyor.analysis.useDefaultRulesets": {
           "type": "boolean",
           "default": true,
           "description": "Whether analysis should use the default rulesets, included in the extension",
           "scope": "window",
-          "order": 4
+          "order": 3
         },
         "konveyor.analysis.customRules": {
           "type": "array",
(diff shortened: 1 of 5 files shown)


vscode/package.json
```diff
@@ -466,53 +453,39 @@
...
-          "default": "0.0.0.0:8080",
-          "description": "Kai Backend URL",
-          "scope": "window"
+          "scope": "window",
+          "order": 7
         },
         "konveyor.kai.providerName": {
           "type": "string",
           "default": "ChatIBMGenAI",
           "description": "URL for the Kai solution server",
```
dymurray, line 486: This description isn't right yeah? Also, I'm told IBM model doesn't work anymore; can we change the default?

sjd78: Ah this stuff (the kai configurations) will be pulled out to an external file in followup PR later today/tomorrow.

sjd78: But I'll read the descriptions over again and make some adjustments.

vscode/src/utilities/configuration.ts
```diff
@@ -13,18 +14,21 @@ export function getConfigKaiRpcServerPath(): string {
   return getConfigValue<string>("kaiRpcServerPath") || "";
 }
 
-export function getConfigLogLevel(): string {
-  return getConfigValue<string>("logLevel") || "debug";
+export function getConfigLogLevel(): ServerLogLevels {
+  return getConfigValue<ServerLogLevels>("logLevel") || "DEBUG";
 }
 
+// TODO: Is this used?
```
djzager, line 21: I think this can be in the follow-up but we should absolutely be passing this onto `kai` either via their cli arguments or in the initialize call.

I do not think users should typically be messing with it but we should keep them as configurable.

vscode/src/client/analyzerClient.ts
```diff
@@ -246,34 +249,33 @@ export class AnalyzerClient {
...
+
+      // TODO: Once konveyor/kai#547 is resolved, multiple bundles can be supported
+      // jdt.ls bundles
+      analyzerLspJavaBundlePath: this.assetPaths.jdtlsBundleJars[0],
+
+      analyzerLspDepLabelsPath: this.assetPaths.openSourceLabelsFile,
 
-      analyzer_lsp_rules_path: this.getRulesetsPath(),
+      // TODO: Do we need to include `fernFlowerPath` to support the java decompiler?
+      // analyzerLspFernFlowerPath: this.assetPaths.fernFlowerPath,
```
djzager, line 278: This makes me want to know two things.
1. Is the decompiler actually included in our assets? If it is we should just pass it on to kai. If not then, 👇 
1. @shawn-hurley or @pranavgaikwad are we doing any decompilation yet? My guess is no. If we don't have the asset and it isn't useful to us right now. Make an issue and we'll do it later.

djzager (approved):
My understanding is that this brings us inline with https://github.com/konveyor/kai/pull/526 . Stealing what Jonah did in #190 😁 . If that is true, this LGTM with the understanding the larger ✨ will come when we move the kai config to it's own file.

vscode/src/utilities/configuration.ts
sjd78: There are no cli args or initialize args that support those settings else I would have passed them along.

vscode/src/client/analyzerClient.ts
sjd78: The provider support a cli argument of `fernFlowerPath` but that is not currently available in the analyzer rpc or the pass-through code.

The jar is currently in the repo here: https://github.com/konveyor/editor-extensions/blob/461e9ce7d7268137a2d2a45859b85c877019588d/assets/fernflower/fernflower.jar

If such a time happens where it is needed, it is easy to pass along.

vscode/src/utilities/configuration.ts
sjd78: I'll update descriptions and comments so they make more sense.

sjd78: Created https://github.com/konveyor/kai/issues/550 and I'll reference those in the code.

vscode/package.json
sjd78: Updated.

djzager (dismissed):

djzager (approved):

Issue #178, opened 2025-01-08
Canceling fix generation doesn't do anything


pranavgaikwad:
When you press the Cancel button when generating a fix, nothing happens.

<img width="881" alt="Image" src="https://github.com/user-attachments/assets/d9bda599-a946-4f48-af0b-332976c8b672" />

sjd78:
### Some analysis 
This button is display on the "Konveyor Analysis View".  The webview will send a `CANCEL_SOLUTION` webview messages, but the message in not current handled in the extension (`vscode/src/webviewMessageHandler.ts`).

The message will need to be implemented.

The `getSolution` implementation doesn't currently include an abort signal / option:
https://github.com/konveyor/editor-extensions/blob/6bdca4cc8318197f86d0d0a1e4b4da8a5563d7b6/vscode/src/client/analyzerClient.ts#L486-L543

### Proper fix
Fixing this properly would require implementing #149 (which would require https://github.com/konveyor/kai/issues/525)

### For now
Simply remove the Cancel button all together until #149 is working.

PR #204, opened 2025-01-16
:sparkles:  Resolution page v2

rszwajko:
Depends on: #195 

Changes:
1. show Resolutions page from Issue tree view via conext button
2. track solution state and collect log messages
3. use chat-like layout
4. support Markdown
5. support light/dark mode


rszwajko:
## Resolution page in light mode
Screenshot from 2025-01-16 23-54-15
Screenshot from 2025-01-17 00-04-27

## Analysis page in light mode with markdown
Screenshot from 2025-01-16 23-53-38

## Resolution page in dark mode
Screenshot from 2025-01-17 00-08-05

## Opening Resolution page from Issue tree view
Screenshot from 2025-01-17 00-08-34

rszwajko:
## Updated styles for file changes

Screenshot from 2025-01-17 15-14-33
Screenshot from 2025-01-17 15-14-06

djzager (commented):
This is really awesome to see! Nice work.

The only issue I see is in Dark Mode, the white text on blue label is hard to read. I'm not sure if the right way to fix it is to change what "blue" is in dark mode, or to change the color on the label when in dark mode.

rszwajko:
@djzager 
> The only issue I see is in Dark Mode, the white text on blue label is hard to read
>
yep.  check the latest commit - I've switched to black.

ibolton336:
Noticing an issue with styling. 

Screenshot 2025-01-20 at 12 50 18 PM

The page component is no longer center aligned and a large space to left is visible when viewing analysis results.

rszwajko:
@sjdy78 @ibolton336 
The layout problems seems related to missing sidebar - after providing a dummy hidden sidebar it looks OK without any tweaks.
Additionally in the latest commit:
1.  I've fixed double scrollbars by moving "Run analysis" button to the masthead  and removing manual height calculation.
2. Removed "Cancel" button as suggested in #178

rszwajko:
### Scrollbars and buttons in the masthead
Screenshot from 2025-01-20 22-59-50
Screenshot from 2025-01-20 22-55-06

rszwajko:
Force pushed to rebase on top of current HEAD.

webview-ui/src/components/ResolutionsPage.tsx
```diff
@@ -1,169 +1,209 @@
...
-      <Page>
-        <PageSection className="pf-v5-u-px-xl pf-v5-u-py-md">
-          <EmptyState variant="lg" icon={WarningTriangleIcon} titleText="No Active Solutions">
-            <EmptyStateBody>There are no solutions to review at this time.</EmptyStateBody>
-          </EmptyState>
-        </PageSection>
-      </Page>
-    );
-  }
+  console.log("Resolution view state:", {
```
sjd78, line 60: useful to keep this log?

sjd78 (approved):
The screen shots look pretty good.  I expect some followup tweaking as it gets more use, but it is a solid base.  Code structure is reasonable.  I only found one lingering console.log that could be kept if it makes sense.

Fighting to get the vscode themes to line up with PF6 styles will be ongoing.  Looks a good at least 90% of the way there and that feels good enough at this point.

webview-ui/src/components/ResolutionsPage.tsx
rszwajko: Yes. Unfortunately the logic got complex in order to provide chat-like experience. The extra flags may be hard to debug - especially that LLM response/errors might not be easy to reproduce.

PR #205, opened 2025-01-17
:sparkles: Add a settings file to configure the model provider

sjd78:
Resolves: #184

Setup a yaml file to hold environment variables and model provider settings to be use while starting and initializing the `kai-rpc-server`.

Summary of changes:
  - At extension activation, all directories that are needed by the extension are verified to exist and created if they don't

  - If a `provider-settings.yaml` does not exist in the extension's global storage, copy a sample file in place

  - The sample `provider-settings.yaml` uses a reference to select which models block is linked to active and therefore used at runtime

  - Access to all of the extensions paths have been collected into `vscode/src/paths.ts`

  - A new command titled "Open the GenAI model provider configuration file" should be used to open the settings file for editing

<!--
## PR Title Prefix

Every **PR Title**


vscode/src/commands.ts
```diff
@@ -201,21 +147,9 @@ const commandsMap: (state: ExtensionState) => {
...
-      if (newKey === undefined) {
-        window.showInformationMessage("No GENAI_KEY entered. Configuration cancelled.");
-        return;
-      }
-
-      await updateGenAiKey(state.extensionContext, newKey);
-      window.showInformationMessage("GENAI_KEY updated successfully!");
+    "konveyor.openModelProviderSettings": async () => {
+      const settingsDocument = await workspace.openTextDocument(paths.settingsYaml);
+      window.showTextDocument(settingsDocument);
```
rszwajko, line 152: I guess we don't need to `await` since method is async but explicit `return` would be more readable.

vscode/src/paths.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  if (needsCreate) {
+    vscode.workspace.fs.createDirectory(joined);
+  }
+  return joined;
+}
+
+export async function ensurePaths(context: vscode.ExtensionContext): Promise<ExtensionPaths> {
+  const firstWorkspace = vscode.workspace.workspaceFolders?.[0];
```
rszwajko, line 46: quick search for `workspace.workspaceFolders?.[0]` reveals 3 more places - it would be nice to switch them to `paths`

vscode/src/paths.ts
```diff
@@ -0,0 +1,76 @@
...
+    const stat = await vscode.workspace.fs.stat(joined);
+    if (stat.type & vscode.FileType.Directory) {
+      needsCreate = false;
+    }
+  } catch {
+    needsCreate = true;
+  }
+
+  if (needsCreate) {
+    vscode.workspace.fs.createDirectory(joined);
```
rszwajko, line 40: `await` ?

vscode/src/paths.ts
```diff
@@ -0,0 +1,76 @@
...
+  const globalScope = context.globalStorageUri;
+  const workspaceScope = context.storageUri!;
+  const workspaceRepoScope = vscode.Uri.joinPath(firstWorkspace.uri, ".vscode");
+  const extResources = vscode.Uri.joinPath(context.extensionUri, "resources");
+  const settingsYaml = vscode.Uri.joinPath(globalScope, "settings", "provider-settings.yaml");
+
+  paths = {
+    extResources,
+    workspaceRepo: firstWorkspace.uri,
+    data: await ensureDirectory(workspaceRepoScope, "konveyor"),
```
rszwajko, line 60: `Promise.all` ?

vscode/src/paths.ts
```diff
@@ -0,0 +1,76 @@
...
+    settings: await ensureDirectory(globalScope, "settings"),
+    settingsYaml,
+    serverCwd: await ensureDirectory(workspaceScope, "kai-rpc-server"),
+    serverLogs: await ensureDirectory(workspaceRepoScope, "konveyor-logs"),
+  };
+  fsPaths = Object.fromEntries(Object.entries(paths).map(([key, uri]) => [key, uri.fsPath]));
+  return paths;
+}
+
+export let paths: ExtensionPaths = new Proxy({} as ExtensionPaths, {
```
rszwajko, line 70: I feel uneasy with mutable export - it might hit us in some non standard cases.
How about returning a constant reference but blocking the access (in the proxy getter) with a mutable private flag?

vscode/resources/sample-provider-settings.yaml
```diff
@@ -0,0 +1,40 @@
+---
+# This is a sample settings file with a few common providers and models defined
+
+environment:
+  ALWAYS_APPLIED_KEY: "envvar to be set regardless of which model is active"
+
+models:
+  OpenAI: &active
```
rszwajko, line 8: IMHO it's more natural to store the active model in the standard settings. If we choose to store it here then we should provide a way to show that in the settings.

vscode/resources/sample-provider-settings.yaml
```diff
@@ -0,0 +1,40 @@
...
+    environment:
+      OPENAI_API_KEY: "key"
+    provider: "ChatOpenAI"
+    args:
+      model: "gpt-4"
+
+  AmazonBedrock:
+    environment:
+      AWS_ACCESS_KEY_ID: "key id"
+      AWS_SECRET_ACCESS_KEY: "secret"
```
rszwajko, line 18: The file contains both public info that could be part of the repository (i.e. model name) and secrets that should be kept away from workspace and repository. Perhaps we should separate both and merge them in the model provider?

vscode/src/commands.ts
sjd78: Only need an await if the command cares about doing something with the TextEditor that `showTextDocument()` returns.  Here the command only cares about opening the editor.

Why does the command's function need a `return`?

vscode/src/paths.ts
sjd78: None of those calls are long running.  I don't think doing them in parallel would help with extension activation time.

vscode/resources/sample-provider-settings.yaml
sjd78: The env vars needed for a model vary enough that keeping the configs all in one file makes the most sense to me.  Feels better to keep everything needed to configure the model in one place.

If not, then the yaml file should really just be a env file and the rest of the model config arguments would be normal settings.json configurations.

vscode/src/paths.ts
sjd78: Yeah I went back and forth about this one.  But there isn't a really good way to do this since the extension needs to be activated to get a context to fill out the paths data properly.

This Proxy only exists to force use of the paths before activation to fail with a reasonable message.

vscode/resources/sample-provider-settings.yaml
sjd78: Lots of variations can be done.

This I'm working off my meeting notes here: https://github.com/konveyor/editor-extensions/issues/184#issuecomment-2583077916

At least for now, I understood the goals as:
  - have one place to configure things so it is easy to configure
  - keep that place outside of the repo so sensitive info won't get committed by accident
  - bonus points if that place can have comments and examples

djzager (commented):
Excellent start to this. Only one comment, the rest looks good from what I can see.

vscode/src/paths.ts
sjd78: Maybe better to have it be a getter function that throws if the extension hasn't been activated yet?

vscode/resources/sample-provider-settings.yaml
```diff
@@ -0,0 +1,40 @@
+---
+# This is a sample settings file with a few common providers and models defined
+
+environment:
+  ALWAYS_APPLIED_KEY: "envvar to be set regardless of which model is active"
+
+models:
```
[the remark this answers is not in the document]
sjd78: The `active` element flows through to our `KaiConfigModels` type.  The `models` element is ignored.  The fancy "&active" anchor is what picks the model to be active with yaml native references.

But a schema for the yaml file would be nice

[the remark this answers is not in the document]
sjd78: I'll add a comment describe the anchor/link stuff at the very least

vscode/src/commands.ts
rszwajko: hmm ...I need to read more about this one.

vscode/src/paths.ts
rszwajko: I had similar thoughts when creating proxy for the `state.data` object in https://github.com/konveyor/editor-extensions/blob/8eb892794747dec1f8bda81bd3b18539895e3614/vscode/src/extension.ts#L57
The reference is constant and the getter logic is delegated.

vscode/resources/sample-provider-settings.yaml
rszwajko: My point that user should be able to check what model is being used in the settings. Actually this should be the most important and visible setting. The details of the models might be configured it in the external file (one ore many). 
The other point is - what happens if config file is shared between workspaces?

djzager: I have to agree with @sjd78 that having these values (models in a provider-settings.yaml and "which model?" setting in settings.json) would be more confusing than having them together.

> Actually this should be the most important and visible setting.

I'm not certain I fully grok what you mean here. I would argue this could be accomplished by the user knowing **exactly** where that value is defined. I imagine a scenario where a user has no configuration, he must first define the provider in the provider-settings.yaml and then he must declare he wants to use that provider in settings.json?

> What happens if config file is shared between workspaces?

Assuming you mean the provider-settings.yaml, it is using the globalStorageUri (line 55 in paths.ts).

vscode/src/paths.ts
sjd78: Converted the paths export from a variable to a function that returns the reference.  Much less magical = easier to understand.

vscode/resources/sample-provider-settings.yaml
sjd78: For 0.0.5, we'll leave the file and handling as-is.

Going forward, there are 2 things that could be added:
  1. Have a substitution from the config file to protected storage so an api key doesn't live as clear text at rest.
  2. Add a new config key to the extension's configuration allowing selection of either the "active" anchor/link block or a named block from the config file to allow selection of the model via the settings.json config file.

rszwajko: I'm OK with leaving it as-it-is  - those concerned about privacy may still provide secrets as env var the usual way.

rszwajko (approved):
Looks good!

PR #206, opened 2025-01-17
:seedling: Fix typing on `Array.filter(Boolean)` for the project

sjd78:
Use a new workspace project `extra-types` that installs to `@types/extra-types` as a way to make `Array.filter(boolean)` have the correct typing across all of the ts projects in the repo.  Typescript "makes visible" all typing from packages installed under `node_modules/@types`.  Put our override there and no extra configurations needed.  Magic.


rszwajko (approved):
Interesting! Thanks for sharing!

ibolton336 (approved):

djzager (approved):

Issue #187, opened 2025-01-09
[BUG] The RPC server can be started multiple times, and the processes won't stop even after closing VS Code


abrugaro:
The "Start Server" button becomes re-enabled before the server has fully started, so if the server takes longer than expected to start, or if the user clicks quickly while the server is still spawning, another instance is triggered. The previous instance remains running, and the processes won't stop even if the user closes VS Code

Video with steps to reproduce:

https://github.com/user-attachments/assets/a649ae4f-77ed-4abc-80b8-bbb0384be8c3

PR #210, opened 2025-01-17
:bug: Add initialization state tracking to prevent spawning duplicate 

ibolton336:
Resolves https://github.com/konveyor/editor-extensions/issues/187

- Update fireStateChange -> fireServerStateChange for clarity
- Add additional server state for initialization. This is purely to handle the unhandled started but not initialized case that can arise for now. 
TODO: Handle readyToInitialize case. Comment out the fire state event until this is ready to be handled.

### DIFF
--- a/vscode/src/client/analyzerClient.ts
+++ b/vscode/src/client/analyzerClient.ts
@@ -43,7 +43,7 @@ export class AnalyzerClient {
 
   private outputChannel: vscode.OutputChannel;
   private assetPaths: AssetPaths;
-  private fireStateChange: (state: ServerState) => void;
+  private fireServerStateChange: (state: ServerState) => void;
   private fireAnalysisStateChange: (flag: boolean) => void;
   private fireSolutionStateChange: (state: SolutionState, message?: string, scope?: Scope) => void;
 
@@ -54,10 +54,11 @@ export class AnalyzerClient {
     mutateExtensionData: (recipe: (draft: ExtensionData) => void) => void,
     private getExtStateData: () => Immutable<ExtensionData>,
   ) {
-    this.fireStateChange = (state: ServerState) =>
+    this.fireServerStateChange = (state: ServerState) =>
       mutateExtensionData((draft) => {
         draft.serverState = state;
         draft.isStartingServer = state === "starting";
+        draft.isInitializingServer = state === "initializing";
       });
     this.fireAnalysisStateChange = (flag: boolean) =>
       mutateExtensionData((draft) => {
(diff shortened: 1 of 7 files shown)


vscode/src/client/analyzerClient.ts
```diff
@@ -109,21 +110,29 @@ export class AnalyzerClient {
     }
 
     this.outputChannel.appendLine(`Starting the kai rpc server ...`);
-    this.fireStateChange("starting");
+
+    if (this.isServerRunning()) {
```
sjd78, line 114: This is kinda the same check as the TODO comment above.  Should we keep those check state TODOs around?

sjd78 (commented):
Looks reasonable.  One question inline.

I usually prefer to check state enums/strings instead of keeping a list of state boolean flags, but that's not a deal breaker.

vscode/src/client/analyzerClient.ts
ibolton336: Ah right yeah this probably isn't needed. Will remove.

vscode/src/client/analyzerClient.ts
```diff
@@ -109,21 +110,23 @@ export class AnalyzerClient {
...
         this.outputChannel.appendLine(`kai rpc server exited [signal: ${signal}, code: ${code}]`);
-        this.fireStateChange("stopped");
+        this.fireServerStateChange("stopped");
       });
 
       this.kaiRpcServer = kaiRpcServer;
       this.outputChannel.appendLine(`kai rpc server successfully started [pid: ${pid}]`);
-      this.fireStateChange("readyToInitialize");
+      //TODO: Handle the case where the server is not ready to initialize
+      // this.fireServerStateChange("readyToInitialize");
```
sjd78, line 126: /Have you seen the server state change to readyToInitialize in error?  `startProcessAndLogStderr()` should not successfully return until the server reports ready.  Is it logging "waited 10000ms for the kai rpc server to be ready, continuing anyway"?

If yes, then this timeout block may need some extra work:
https://github.com/konveyor/editor-extensions/blob/8eb892794747dec1f8bda81bd3b18539895e3614/vscode/src/client/analyzerClient.ts#L203-L206

At the very least, let's move this TODO comment to that location.

sjd78 (commented):
Only a question about a comment

sjd78 (approved):

sjd78:
Resolving the conflict and this PR should be good to go.

sjd78 (approved):
LGTM

PR #222, opened 2025-01-20
:ghost: Enable use of a proposed api in dev mode

sjd78:
In development mode, to avoid errors being displayed about proposed API use:
  - Add the proposed API to the extension's `package.json` so it can be activated from the vscode cli
  - Add the cli switch to the dev launcher.json config

Note: According to [1], an extension using a proposed api cannot be published.  More exploration is needed to see how we can support using the proposed api via `.vsix` install but also create a package that can be published to the extension store.

[1] - https://code.visualstudio.com/api/advanced-topics/using-proposed-api


rszwajko:
@jwmatthews @sjd78 
1. I could not find a date/plan for publishing the API needed for diff editor to be fully functional - my question in the [November Iteration Plan](https://github.com/microsoft/vscode/issues/232927#issuecomment-2483660753) was not answered (as all others)
2. there is also https://github.com/microsoft/vscode/issues/234097 - a bug that impacts usability of diff editor. I've submitted  and provided an example to reproduce the problem but there is also no clear fix date.

Considering the above - perhaps we should switch to merge editor as the default?

sjd78:
> @jwmatthews @sjd78
> 
>     1. I could not find a date/plan for publishing the API needed for diff editor to be fully functional - my question in the [November Iteration Plan](https://github.com/microsoft/vscode/issues/232927#issuecomment-2483660753) was not answered (as all others)
> 
>     2. there is also [vscode.workspace.applyEdit opens a duplicated editor for left-side edits in diff editor microsoft/vscode#234097](https://github.com/microsoft/vscode/issues/234097) - a bug that impacts usability of diff editor. I've submitted  and provided an example to reproduce the problem but there is also no clear fix date.
> 
> 
> Considering the above - perhaps we should switch to merge editor as the default?

Probably best so the extension can be published...  We'll probably needs some processing in the copy scripts to filter out commands that hook the proposed APIs so the built extension is "clean".

sjd78:
@konveyor/kai-ide-reviewers -- We should put this in regardless and deal with publishability to a marketplace in the future.  This PR will just make life a little nicer for us right now.

djzager (approved):

PR #261, opened 2025-01-23
:ghost: Change sample to use 4o model

ibolton336:


dymurray (approved):

sjd78 (approved):
FWIW, the sample file is only pushed to the actual `provider-settings.yaml` when the latter doesn't exist.  Maybe a command to "reset provider settings to extension defaults" would be useful?  That command should not just overwrite, but move the existing file to `${basename}.YYYYMMDD.yaml` kind of thing...

sjd78:
See #264

Issue #221, opened 2025-01-20
[BUG] Text Visibility Issue in Light Theme


midays:
**Description:**

In the Konveyor Analysis View page, the "No Violations Found" text changes color based on the theme. When the user switches to the light theme, the text changes to a dark color while the rest of the UI elements, including the background, remain dark-themed. This makes the text invisible or difficult to read.

**Steps to Reproduce:**

1. Open the Konveyor Analysis View page.
2. Switch to the light theme.
3. Observe the visibility of the "No Violations Found" text.

**Expected Behavior:**

The analysis view tab should adapt to the selected theme, ensuring proper contrast and readability.

**Actual Behavior:**

The text becomes dark in the light theme, blending with the dark background.

**Environment:**

OS: Linux/fedora
PLugin Version: 0.0.4

https://github.com/user-attachments/assets/135d409b-fc1f-4bf5-b445-9d29c311f97c

rszwajko:
In #204 we will provide better support for dark/light modes.

PR #303, opened 2025-01-27
:bug: Use vscode theme for color and bg color where possible

ibolton336:
Resolves https://github.com/konveyor/editor-extensions/issues/221


webview-ui/src/components/ResolutionsPage/ResponseWrapper.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { FC } from "react";
+import { Label, FlexItem } from "@patternfly/react-core";
+
+interface ResponseWrapperProps {
+  type: "sent" | "received";
+  children: React.ReactNode;
+  className?: string;
+}
+
+export const ResponseWrapper: FC<ResponseWrapperProps> = ({ type, children, className = "" }) => {
```
sjd78, line 10: Why not just have 2 components:
  - SentMessage
  - ResponseMessage

I have it in the back of my mind to end up with a few kinds of response/activity messages and it would be nice to have multiple render components based on those types.   (ResponseStringMessage, ResponseMarkdownMessage, ResponseFancyPantsFormattedMessage, ...)

webview-ui/src/index.css
```diff
@@ -1,4 +1,20 @@
...
+
+:root {
+  --pf-v6-c-card--BackgroundColor: var(--vscode-editor-background) !important;
+  --pf-t--global--background--color--primary--default: var(--vscode-editor-background) !important;
+  --pf-v6-c-content--Color: var(--vscode-editor-foreground) !important;
+  --pf-v6-c-page--BackgroundColor: var(--vscode-editor-background) !important;
+  --pf-v6-c-page__main-container--BackgroundColor: var(--vscode-editor-background) !important;
+  --pf-t--global--text--color--link--100: #3794ff; /* VS Code textLink.foreground */
+  --pf-t--global--color--status--danger--100: #f48771; /* VS Code errorForeground */
+}
```
sjd78, line 17: Ah ha, that's what I expected to see more of...overrides of the PF color/font variables.

sjd78 (commented):
Visually it looks better.

Structurally looks good, with maybe a tweak to the response wrapper component.  Even that could be done in the future as we want to render responses/notifications with more structure.

I did expect to see more --pf variables being overridden to control the colors, but that's not even a problem.

webview-ui/src/components/ResolutionsPage/SentResponse.tsx
djzager, line 1: My only nitpick on this is the naming. This is referring to a message we have sent...right? Maybe a less confusing set of component names would be `SentMessage` `ReceivedMessage`, to me those would be less confusing.

djzager (commented):
One non-blocking comment. Let me know what you think and I'll re-review.

webview-ui/src/components/ResolutionsPage/SentResponse.tsx
sjd78: Those are better names for sure.  I just saw "Sent" and "Received" and went with it.

webview-ui/src/components/ResolutionsPage/ResolutionsPage.tsx
```diff
@@ -78,23 +76,14 @@ const ResolutionPage: React.FC = () => {
...
               grow={{ default: "grow" }}
               alignItems={{ default: "alignItemsFlexEnd" }}
-              justifyContent={{ default: "justifyContentFlexEnd" }}
             >
-              <FlexItem>
-                <YellowLabel>Here is the scope of what I would like you to fix:</YellowLabel>
-              </FlexItem>
+              <SentResponse>Here is the scope of what I would like you to fix:</SentResponse>
               <FlexItem className="chat-card-container">
                 <ChatCard color="yellow">
```
rszwajko, line 88: ChatCard is no longer yellow but the rest of user triggered messages is still yellow. We should be consistent here. If keeping the yellow background (even) for light theme is complicated then lets drop it and use system background. Fortunately the blue background seems universal so we maintain visual difference with the server side messages.

webview-ui/src/components/ResolutionsPage/ResolutionsPage.tsx
```diff
@@ -104,73 +93,52 @@ const ResolutionPage: React.FC = () => {
...
-                <ChatCard color="blue">
-                  <FileChanges
-                    changes={getRemainingFiles()}
-                    onFileClick={handleFileClick}
-                    onApplyFix={handleAcceptClick}
-                    onRejectChanges={handleRejectClick}
-                  />
-                </ChatCard>
-              </FlexItem>
+              <ReceivedResponse>
```
rszwajko, line 115: Bigger objects look really bad inside Label. Actually I consider a Label a temporary approach until we figure out a better way to display the messages.

webview-ui/src/components/ResolutionsPage/resolutionsPage.css
```diff
@@ -0,0 +1,72 @@
...
+  max-width: 80%;
+  border-radius: 20px;
+  padding: 12px;
+  margin: 4px 0;
+  box-shadow: 0 1px 2px rgba(0, 0, 0, 0.1);
+}
+
+/* Sent message (yellow) */
+.pf-v6-c-card.pf-m-yellow {
+  border-top-right-radius: 4px;
```
rszwajko, line 21: why top right corner has different radius?

rszwajko (commented):
Keeping VS Code and PF styles will be a though job!
Second thought: we need to rethink how we differentiate server vs user messages: currently we use alignment (left/right) and color coding(yellow/blue or native/blue). Moving our color coding entirely to native/blue would help for sure but perhaps we could do more:
1. modify color coding: i.e. drop backgrounds and use borders and/or icons 
2. use text label ("name") to identify server/AI (following  PF [Chatbot guidelines ](https://www.patternfly.org/patternfly-ai/conversation-design/#ai-disclosure))

sjd78 (approved):
The resolution response view messages will need some additional tweaking, but that's best left for a followup PR.

sjd78:
> The resolution response view messages will need some additional tweaking, but that's best left for a followup PR.

#352

PR #310, opened 2025-01-28
:sparkles: add config options for cache_dir and trace

pranavgaikwad:
This adds the ability for users to generate new cached data and also adds an option to generate the LLM traces.


djzager (approved):
lgtm

PR #315, opened 2025-01-29
:sparkles: Added preliminary notifications

JonahSussman:


vscode/src/client/analyzerClient.ts
```diff
@@ -140,6 +141,15 @@ export class AnalyzerClient {
...
       new rpc.StreamMessageWriter(this.kaiRpcServer.stdin),
     );
+
+    this.rpcConnection.onNotification("my_progress", (params) => {
+      if (params.kind === "SimpleChatMessage") {
+        this.fireSolutionStateChange("sent", params.value.message);
+      } else {
+        this.fireSolutionStateChange("sent", JSON.stringify(params));
+      }
+    });
```
sjd78, line 151: Could add a little bit of protection by only doing something if the solution state is already `"sent"`...
```suggestion
    this.rpcConnection.onNotification("my_progress", (params) => {
      if (this.getExtStateData().solutionState !== "sent") {
        return;
      }
      if (params.kind === "SimpleChatMessage") {
        this.fireSolutionStateChange("sent", params.value.message);
      } else {
        this.fireSolutionStateChange("sent", JSON.stringify(params));
      }
    });
```

sjd78 (commented):
One non-blocking comment

vscode/src/client/analyzerClient.ts
JonahSussman: Perhaps I'm misunderstanding. Will this ever block a chat message from being sent?

djzager: no. He was saying that his comment on your PR itself was non-blocking.

sjd78: I know this function will get some upgrades in future, so it is ok as-is at the moment.

sjd78 (approved):

PR #193, opened 2025-01-13
:ghost: Update the shared workspace to build to esm and commonjs

ibolton336:
Resolves https://github.com/konveyor/editor-extensions/issues/192

Vite Library Mode Integration:
> Enables efficient bundling of the shared package using Vite, producing both CommonJS (CJS) and ES Module (ESM) outputs. This setup ensures compatibility with modern and legacy environments for seamless use in both the VSCode extension and webview packages.

Summary of changes:
  - drop typescript composite project structures
  - build `shared` with vite in library mode 
  - fixup the dev scripts so the `shared` build has a chance to run before the other packages go looking for its files
  - all workspaces now launch in dev or build watch mode - changes in any of them will result in recompiles

PR #317, opened 2025-01-29
:sparkles: On analysis webview, add group by files or issues

ibolton336:
This PR refactors how incidents and violations are processed within the IDE to better align with the backend representation. The goal is to ensure that all incidents sent to kai contain necessary metadata for accurate solution generation. 

✅ Introduces EnhancedIncident and EnhancedViolation types

These types extend the base Incident and Violation types with additional metadata required by kai.
Ensures consistency between what we receive from analysis and what we send to kai.

✅ Refactors solution request logic

getSolution now strictly uses EnhancedIncident, removing the need for separate mutation steps.
Violation metadata (ruleset_name, violation_name, etc.) is now embedded directly in the incident data.
Removes redundant violation parameters from solution-related commands and function calls.

✅ Updates Webview & VS Code Extension state

Introduces a new enhancedIncidents field in ExtensionData to store enriched incidents.
Adjusts how violations are computed using useViolations(), simplifying data processing in the webview. There is probably some more work to do around weeding out unneeded data hanging around after the introduction of enhancedIncidents.


webview-ui/src/utils/transformation.ts
```diff
@@ -31,5 +31,25 @@ export const groupIncidentsByMsg = (
         acc[msg].push([uri, incident]);
         return acc;
       },
-      {} as { [msg: string]: [string, Incident][] },
+      {} as { [msg: string]: [string, EnhancedIncident][] },
     );
+
+export function enhanceIncidents(
```
djzager, line 37: I wonder, if instead of having this function, we updated the loadRulesets to return an `EnhancedIncident[]`. At that time we have all of the information we need to properly construct these incidents. If we pull that thread to completion, we can eliminate the need for dragging rulesets and violations all around our codebase.

webview-ui/src/utils/transformation.ts
```diff
@@ -31,5 +31,25 @@ export const groupIncidentsByMsg = (
...
+
+  return incidents.map((incident) => ({
+    ...incident,
+    violationId: violation.id,
+    violationDescription: violation.description,
+    rulesetName: violation.rulesetName,
+    violationName: violation.violationName,
+    uri: incident.uri,
+    message: incident.message,
+    severity: incident.severity,
```
djzager, line 53: This builds on what I said about considering moving this into `loadRulesets` where we have the rulesets + violations + incidents all available to us, but we must make sure that we fill in all of this correctly to give `kai` the best chance of providing meaningful solutions [here](https://github.com/konveyor/kai/blob/de6ca2ea455b9960dae5cded53112e18b1d24228/kai/analyzer_types.py#L89-L100).

```
    # copied it all here for ref
    ruleset_name: str
    ruleset_description: Optional[str] = None

    violation_name: str
    violation_description: Optional[str] = None
    violation_category: Category = Category.POTENTIAL
    violation_labels: list[str] = []
```

djzager (commented):
I am hopeful that we can do all this transformation at the time we load the rulesets and save ourselves a lot of headache.

shared/src/types/types.ts
```diff
@@ -27,16 +27,31 @@ export interface Violation {
...
 }
 
-export type ViolationWithID = Violation & {
+export type EnhancedViolation = Violation & {
   id: string;
+  rulesetName?: string;
+  violationName?: string;
 };
 
+export interface EnhancedIncident extends Incident {
```
shawn-hurley, line 36: Should this just be one type Incident? do we ever want non enhance incidents?

shawn-hurley (commented):
Just one question, I don't know if it make sense or not

shared/src/types/types.ts
```diff
@@ -27,16 +27,31 @@ export interface Violation {
   effort?: number;
 }
 
-export type ViolationWithID = Violation & {
+export type EnhancedViolation = Violation & {
   id: string;
+  rulesetName?: string;
+  violationName?: string;
 };
```
djzager, line 34: I guess this is because we need it already...why not put the extra bits in there?

shared/src/types/types.ts
djzager: I think this structure comes into play when we unmarshal the analysis response from kai. I'm not certain if only having the enhanced would cause any problems.

I rate my conviction on this...meh...but I would prefer to keep them distinct with a comment that Incident is what we get from analysis and EnhancedIncident is what we send to kai.

vscode/package.json
```diff
@@ -32,10 +32,6 @@
     "node": ">=22.9.0",
     "npm": "^9.5.0 || ^10.5.0"
   },
-  "activationEvents": [
```
djzager, line 35: I'm leaving a comment on this so we can talk about this before this gets merged.

webview-ui/src/components/IncidentTable/IncidentTableGroup.tsx
```diff
@@ -10,24 +9,34 @@ export const IncidentTableGroup = ({
   workspaceRoot,
   incidents,
 }: {
-  violation?: Violation;
-  onIncidentSelect: (incident: Incident) => void;
-  onGetSolution?: (incidents: Incident[], violation: Violation) => void;
+  violation?: EnhancedViolation;
```
djzager, line 12: Do we still need this?

djzager (commented):
This is much more in line with what I had in mind. I'm leaving the comments so I can play around with it in the morning.

webview-ui/src/utils/transformation.ts
```diff
@@ -17,19 +17,3 @@ export const sanitizeIncidents = (incidents: Incident[]): Incident[] =>
       // expect 1-based numbering (vscode.Position is zero-based)
       lineNumber: Number.isInteger(it.lineNumber) && it.lineNumber! > 0 ? it.lineNumber : 1,
     }));
-
-export const groupIncidentsByMsg = (
```
rszwajko, line 21: This was a temp placeholder until we will be able to share code via `shared` project. This was fixed in https://github.com/konveyor/editor-extensions/pull/193 . 
Now we should be able to use this functionality from https://github.com/konveyor/editor-extensions/blob/2f4a371688f41199a09318adc22baa9f7cd7c1f9/shared/src/transformation.ts#L21
This is important to keep the groupings in Issue tree view (and possibly other components) look exactly the same as the groupings used in the webview.

shared/src/types/types.ts
rszwajko: The enhanced incidents seems the IDE version of https://github.com/konveyor/kai/blob/82e195916be14eddd08c4e2bfb69afc0880edfcb/kai/analyzer_types.py#L89-L106

@ibolton336  please add a comment pointing to that type to keep both aligned in the future

rszwajko (commented):
Looks like we are aligning our types with the backend https://github.com/konveyor/kai/blob/de6ca2ea455b9960dae5cded53112e18b1d24228/kai/analyzer_types.py#L89

@ibolton336 please update the PR title/description

shared/src/types/types.ts
rszwajko: @JonahSussman please take a look if the flattening in the IDE works the same as in the backend.

sjd78 (commented):
Terminology question.  The api calls things violations and incidents.  On the UI we've called them issues and incidents.

Can we stick with using "Issue" instead of "Violation" to be consistent with Konveyor UI?

vscode/package.json
sjd78: This would be better in another small PR so it is easier to track.

webview-ui/src/components/IncidentTable/IncidentTable.tsx
```diff
@@ -44,14 +50,20 @@ export const IncidentTable: FC<IncidentTableProps> = ({
               ? {
                   hasNoOffset: true,
                   actions: (
-                    <ViolationActionsDropdown
-                      onGetAllSolutions={() => getSolution(incidents)}
-                      fixMessage={
+                    <Button
```
sjd78, line 53: I prefer the kebab for the actions, but we can change the up later.

vscode/package.json
sjd78: Moved to #323

webview-ui/src/utils/transformation.ts
sjd78: Ah, I see what Radek is saying.

Incidents can be grouped by description/message under their Issue.  The issue tree view shows per description/message while the analysis view is back to per Issue.  We can solve this in future.  I'll kick it to an issue so we can have a discussion about it and display things the same way in both views.

ibolton336: +1

ibolton336: https://github.com/konveyor/editor-extensions/issues/325

sjd78: I'll just put some more details in that issue you opened!  :+1:

djzager (approved):
LGTM.

image
image

And the request we are sending (when I request a solution for a file and it's incidents) is correct with all of the relevant ruleset + violation details.

PR #318, opened 2025-01-29
:bug: fail visibly on server start

### LINKED ISSUE #309: Duplicate map keys under models in provider-settings.yaml prevents server from starting with no easy to see error
I copied a provider and forgot to update the name and ended up with:
```
models:
  rhoai-proxy:
  ...
  rhoai-proxy: &active
  ...
```

When trying to start the server it spins for a moment and fails to start, without an obvious error. If you navigate to View>Output and select Konveyor-Analyzer from the drop down it is pretty easy to see the cause, but it would be helpul to show a more obvious error if possible.

djzager:
Fixes #309


djzager:
Uploading Screen Recording 2025-01-29 at 16.10.51.mov…

ibolton336 (approved):

PR #319, opened 2025-01-29
:seedling: Add ability to trace the jsonrpc MessageConnection

sjd78:
To make it easier to work between the jsonrpc server and the extension client, add the ability to trace the messaging to an output channel.

Details:
  - The output channel's name is of the form: `${basename(kai-rpc-server)} message trace`

  - The trace output channel is created when the server is first started, and is independent from the `Konveyor-Analyzer` channel

  - Tracing is controlled by the configuration key `konveyor.logging.traceMessageConnection`. It is `false` by default.

Note: Supports #315

### DIFF
--- a/vscode/src/client/tracer.ts
+++ b/vscode/src/client/tracer.ts
@@ -1,0 +1,65 @@
+// Modified from https://github.com/microsoft/vscode-languageserver-node/blob/main/client/src/common/client.ts
+
+import * as vscode from "vscode";
+import { Logger, ResponseError, Tracer } from "vscode-jsonrpc/node";
+
+function isString(value: any): value is string {
+  return typeof value === "string" || value instanceof String;
+}
+
+function data2String(data: object): string {
+  if (data instanceof ResponseError) {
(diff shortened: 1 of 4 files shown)


sjd78:
When enabled, will look like this:
image

djzager (approved):
lgtm

vscode/src/client/tracer.ts
```diff
@@ -0,0 +1,65 @@
...
+    return data;
+  }
+  return data.toString();
+}
+
+const tracerToLogOutputChannel = (channel: vscode.LogOutputChannel): Tracer => {
+  const logTrace = (message: string, data?: any) => {
+    channel.info(message);
+    if (data) {
+      channel.info(data2String(data));
```
rszwajko, line 31: @sjd78 
why we are not using `logObjectTrace` here?

sjd78: That's just what was in the code I borrowed it from, so I didn't think very hard about it.  :man_shrugging:

PR #323, opened 2025-01-30
:bug: Remove activation events causing stale data to load

ibolton336:
- This code reverts a bug fix https://github.com/konveyor/editor-extensions/pull/275 
- Introduced a new bug causing stale data to be loaded into memfs 
Screenshot 2025-01-31 at 11 10 49 AM

Notice in the Konveyor resolutions pane in bottom left, a solution  from a former run is loaded in when a getSolution request is fired on an unrelated issue. @djzager @sjd78

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -32,10 +32,6 @@
     "node": ">=22.9.0",
     "npm": "^9.5.0 || ^10.5.0"
   },
-  "activationEvents": [
-    "onFileSystem:konveyorMemFs",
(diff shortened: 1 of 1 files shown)


djzager (commented):
I suspect we need to reopen the original issue so we can figure out a way to address it? Or is it a problem with our extension startup that makes these activation events a problem?

It isn't immediately clear the path to resolving both the problem we created by adding activation events AND the issue that caused us to add activation events in the first place.

Issue #252, opened 2025-01-23
The "folder" column in the incidents always displays "file://"


abrugaro:
![Image](https://github.com/user-attachments/assets/2e63b3d4-2131-44f8-9bdc-148aab6b7f76)

Update: The issue is reproducible only for Java files, in `xml` files it works fine, althought the `file://` prefix should be removed

![Image](https://github.com/user-attachments/assets/8ecfd6cf-c6a8-447b-8149-0c144b190fca)

**Env:**
* Windows 11 Pro
* Extension v0.0.5

rszwajko:
@abrugaro @sjd78 
this looks like a regression after configuration changes - in https://github.com/konveyor/editor-extensions/pull/204 you can see that the folder path is relative to the workspace folder. In order to calculate the relative path the webview needs a `workspaceRoot` prop.

abrugaro:
Hi, reopening this as it is still reproducible on 0.0.7.

![Image](https://github.com/user-attachments/assets/fc0237cb-78bb-46f0-8dee-8c6d0fd814b5)

Tested on Windows 11 pro

sjd78:
Closing since the #357 tracks the remaining issue on windows

PR #324, opened 2025-01-30
:bug: Fix Analysis and Resolution file paths relative to workspaceRoot

sjd78:
Incident file path information changed to use Uri values.  Since the `workspaceRoot` being handed to the webviews was fsPath based, the full Uris were being rendered.  Pass the workspaceRoot as a Uri string and get the `relativeDirname()` function working properly again.

Resolves: #252


sjd78:
Before Change:
image

After Change:
image

djzager (approved):

PR #331, opened 2025-01-31
:sparkles: better solution configuration

djzager:


vscode/package.json
```diff
@@ -544,61 +544,32 @@
...
-        "konveyor.kai.getSolutionMaxIterations": {
-          "type": [
-            "number",
-            "null"
+        "konveyor.kai.getSolutionMaxEffort": {
+          "type": "string",
+          "enum": [
+            "Low",
+            "Medium",
+            "Maximum"
```
fabianvf, line 556: Can we do `Maximum (experimental)` just so people know it's a dangerous button

djzager:
image
image

sjd78 (approved):
Looks reasonable so long as having the same settings for a single incident or multiple incidents is ok.

fabianvf (approved):
We've got changes planned for the backend that should make the performance improve enough that we don't need to separate it anymore, plus the message passing that's going in the next release will help a lot too

PR #333, opened 2025-01-31
:seedling: collect-assets refactored to use GitHub API with auth

sjd78:
Refactor the `collect-assets` scripts so every call to the GitHub API can be authorized with the GITHUB_TOKEN envvar.  This will help prevent rate limiting encountered by CI and Release actions.

Summary:
  - Setup `GITHUB_TOKEN` in actions as needed
  - Use `@octokit/core` for GitHub API access as it can default auth and it has lots of nice typing and rich information if errors are encountered
  - GitHub fetch functions are now all in `_github.js`
  - try/catch blocks moved as high as possible, any `await function` call that fails will be passed up

### DIFF
--- a/.github/workflows/ci-repo.yml
+++ b/.github/workflows/ci-repo.yml
@@ -92,6 +92,19 @@ jobs:
         run: |
           npm ci
 
+      #
+      # TODO: To run a proper test, the Kai assets need to be downloaded for the proper place
+      # TODO: For example, this would be nice...
+      # - name: Collect assets needed to run Kai on the current runtime's platform/arch
+      #   env:
+      #     GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      #   run: |
+      #     npm run collect-assets \
+      #       --workflow \
+      #       --branch=main \
+      #       --platform=runtime \
+      #       --arch=runtime
+
       # Run tests on Linux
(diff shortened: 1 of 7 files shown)


savitharaghunathan (commented):
/lgtm

tested the workflow locally and it works. 
Note: when downloading a release `java-deps.zip` is not present
```Asset [java-deps.zip] was not found in GitHub release https://github.com/konveyor/kai/releases/tag/v0.0.7```

do we need to tackle that? 

I havent tested the ci/release workflow. VISIACK for gh actions part of the PR

sjd78:
> /lgtm
> 
> tested the workflow locally and it works. Note: when downloading a release `java-deps.zip` is not present `Asset [java-deps.zip] was not found in GitHub release https://github.com/konveyor/kai/releases/tag/v0.0.7`
> 
> do we need to tackle that?

Yes and no. It’s not a published release asset, it is a published workflow artifact. It needs another conversation at the very least.

but that’s nothing that needs to go in this PR

> 
> I havent tested the ci/release workflow. VISIACK for gh actions part of the PR

those changes are mostly just to get the GitHub token available to the scripts

djzager (commented):
lgtm

djzager (approved):
lgtm...actually approve this time.

Issue #327, opened 2025-01-31
Fail quickly and warn when we see an error like: 'invalid_api_key'


jwmatthews:
Let's ensure that we fail quickly on the case when Kai is having known problems with talking to a service which will not resolve.  We could consider to start this that an API Key error is a good candidate of something that will not resolve on it's own, so we should fail fast when detected.  This will get a little complicated later as we consider potential intermittent issues which could resolve on a retry.  

Below is my reproducer

I uninstalled extension 0.0.6 and installed 0.0.7.
I FORGOT to configure my provider key via `provider-settings.xml`
`provider-settings.xml`
```
models:
  OpenAI: &active
    environment:
      OPENAI_API_KEY: "key"
    provider: "ChatOpenAI"
    args:
      model: "gpt-4o"
```

So I was running with default with no useful key configured.
This is my fault of not configuring the key, yet I assume this will be a common workflow folks hit at first.

I proceeded to run analysis and looked good, so I asked to resolve a fix.
It ran for 12 minutes, and then it said no response was found

<img width="1579" alt="Image" src="https://github.com/user-attachments/assets/689ba1d7-d556-4f04-b912-821bca4c1d28" />

Below is look at the response

```
$ head -n 10 .vscode/konveyor/solution_20250131T144638.json
{
  "encountered_errors": [
    "Error code: 401 - {'error': {'message': 'Incorrect API key provided: key. You can find your API key at https://platform.openai.com/account/api-keys.', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_api_key'}}",

```

djzager:
imo what remains after #335 is complete is on kai to prevent initialize being successful when the provider info specified isn't useful for actually communicating with the provider.

Documented here https://github.com/konveyor/kai/issues/609

PR #335, opened 2025-02-01
:sparkles: Remove key text from `sample-provider-settings.yaml`, add a command to reset

sjd78:
Two changes:
  - Remove the sample key text to shorten any time to fail on solution requests if a user doesn't put the keys in the right place

  - Add a command to allow a user to backup their existing `provider-setting.yaml` file and reset it back to the extension default/sample file

This is a partial fix for #327.

### DIFF
--- a/vscode/package.json
+++ b/vscode/package.json
@@ -39,8 +39,14 @@
   "contributes": {
     "commands": [
       {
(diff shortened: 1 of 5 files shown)


djzager (approved):
lgtm

PR #340, opened 2025-02-03
:bug: expose demoMode

pranavgaikwad:


vscode/package.json
```diff
@@ -594,6 +594,13 @@
           "scope": "window",
           "order": 50
         },
+        "konveyor.kai.demoMode": {
+          "type": "boolean",
+          "default": false,
+          "description": "Whether to run Kai in demo mode i.e. use cached LLM responses",
+          "scope": "window",
+          "order": 50
+        },
```
djzager, line 603: Can you not infer this from a non-null `konveyor.kai.cacheDir`? Why do we need both? Are we sure we should be sticking demo flags directly into our IDE config?

djzager (approved):
LGTM.

I'm leaving the comment on whether this is actually necessary or not because I still don't know that it is. But the strength of my conviction is not enough to not approve.

If we can do without it though, infering it from the cacheDir I personally think that would be better.

vscode/package.json
pranavgaikwad: @djzager there is a usecase where you would want the builtin cache (not setting cacheDir) to be used when in demo mode (setting demoMode). I don't see this being used by users that often, but for our demos we need this kind of thing.

PR #343, opened 2025-02-04
:seedling: Enable mocha testing on webview-ui

sjd78:
Enable mocha tests on `webview-ui` by:
  - moving mocha deps to root `package.json`
  - adding `.mocharc.json` to each project that has unit tests
  - adopt `tsx` instead of `ts-node`

Note: A glob is in place on `webview-ui`'s test script since no unit test currently exist to run.  In that case, it isn't really an error.  In the future, once a test is added, the glob check should be removed so mocha will fail if tests are not found.

Supports: #338


djzager (approved):
LGTM

PR #346, opened 2025-02-04
:seedling: add another effort level before max

djzager:


fabianvf (dismissed):

sjd78 (dismissed):
Only comment is an edge case for config values.  Best to put the check in, but can get by without.

sjd78 (approved):
LGTM

PR #355, opened 2025-02-06
:sparkles: Tell `kai` to reset the analyzer cache when asking for full analysis

shawn-hurley:


shawn-hurley:
This is needed for/by https://github.com/konveyor/kai/pull/623 to work correctly

vscode/src/client/analyzerClient.ts
```diff
@@ -441,8 +441,13 @@ export class AnalyzerClient {
           const requestParams = {
             label_selector: getConfigLabelSelector(),
             included_paths: filePaths?.map((uri) => uri.fsPath),
+            reset: false,
           };
 
+          if (!requestParams.included_paths) {
+            requestParams.reset = true;
+          }
+
```
djzager, line 450:
```suggestion
          const requestParams = {
            label_selector: getConfigLabelSelector(),
            included_paths: filePaths?.map((uri) => uri.fsPath),
            reset: !(filePaths && filePaths.length > 0),
          };
```

djzager (changes requested):
One nit.

vscode/src/client/analyzerClient.ts
```diff
@@ -441,8 +441,8 @@ export class AnalyzerClient {
           const requestParams = {
             label_selector: getConfigLabelSelector(),
             included_paths: filePaths?.map((uri) => uri.fsPath),
+            reset: !(filePaths && filePaths.length > 0),
```
rszwajko, line 444: @shawn-hurley 
we have a command to clear analysis results in IDE. We should either clear the cache from that command or clear it on the next partial analysis. Otherwise discarded results will get resurrected after file save :)

djzager (approved):
LGTM.
I updated the title to better reflect the cache we are talking about.

vscode/src/client/analyzerClient.ts
rszwajko: Yep. I think we should be able to clear it on demand.
However this is not a new problem - I've tested it on older version (before this PR).
We can fix it in follow-up  https://github.com/konveyor/editor-extensions/issues/356

Issue #272, opened 2025-01-24
Stop/kill the kai rpc server when the extension is deactivated / the workspace's vscode instance is closed


sjd78:
When the vscode window is closed, the process itself end.  But if the kai-rpc-server has been started, it needs to be stopped.  We don't want to leave orphaned processes running.

- Check that the extension is deactivated with the vscode window in closed
- Check that deactivation will cause the server to be shutdown and killed

Events that should cause the extension to deactivate and kai-rpc-server to be shutdown/killed:
  - Closing the vscode window for the workspace
  - Disabling the extension
  - Running the "Restart the extension host" command
  - Opening a new folder in the vscode window

PR #359, opened 2025-02-07
:bug: Properly stop the jsonrpc server on extension deactivation

sjd78:
If the jsonrpc server is running when the extension is deactivated, which is usually caused by exiting vscode, stop the server.

Resolves: #272


djzager (approved):
lgtm.

PR #365, opened 2025-02-11
:sparkles: make the incident list sortable

### LINKED ISSUE #358: Allow sorting of Files and/or Issues in Analysis View
As a user from the IDE I would like the ability to sort by Files or Issues.

When viewing my Issues, would be nice to sort alphabetical
Image

Similar with by File
Image

### LINKED ISSUE #369: Rule severity filter on analysis view should be changed to category instead
This should change to match on rule category which is either one of - mandatory, potential or optional

Image

djzager:
* Make the list of incidents sort ascending/descending based on the grouping (issues | files)
* Add new toggle for the sort
* Swap severity filter for category filter that actually exist on incidents.

Fixes #358 
Fixes #369

### DIFF
--- a/webview-ui/src/components/ViolationIncidentsList.tsx
+++ b/webview-ui/src/components/ViolationIncidentsList.tsx
@@ -27,10 +27,17 @@ import {
   ToggleGroup,
   ToggleGroupItem,
 } from "@patternfly/react-core";
-import { WrenchIcon, ListIcon, FileIcon, LayerGroupIcon } from "@patternfly/react-icons";
+import {
+  WrenchIcon,
+  ListIcon,
+  FileIcon,
+  LayerGroupIcon,
+  SortAmountDownIcon,
+  SortAmountUpIcon,
+} from "@patternfly/react-icons";
 import { IncidentTableGroup } from "./IncidentTable";
 import * as path from "path-browserify";
-import { EnhancedIncident, Incident, Severity } from "@editor-extensions/shared";
+import { EnhancedIncident, Incident, Category } from "@editor-extensions/shared";
 
 type GroupByOption = "none" | "file" | "violation";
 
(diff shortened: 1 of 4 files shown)


webview-ui/src/components/ViolationIncidentsList.tsx
```diff
@@ -138,6 +162,48 @@ const ViolationIncidentsList = ({
...
+          isExpanded={isSortOpen}
+        >
+          {isSortAscending ? <SortAmountUpIcon /> : <SortAmountDownIcon />} Sort by{" "}
+          {sortField.charAt(0).toUpperCase() + sortField.slice(1)}
+        </MenuToggle>
+      )}
+      onSelect={() => setIsSortOpen(false)}
+    >
+      <DropdownItem key="none" onClick={() => handleSortSelect("none")}>
+        None
```
ibolton336, line 181: I would change this to "default". Otherwise, the PR

ibolton336 (changes requested):

webview-ui/src/components/ViolationIncidentsList.tsx
```diff
@@ -183,12 +249,22 @@ const ViolationIncidentsList = ({
...
       id,
       label,
       incidents,
     }));
-  }, [enhancedIncidents, searchTerm, filters]);
+
+    if (sortField !== "none") {
+      sortedGroups.sort((a, b) => {
+        const fieldA = a.label.toLowerCase();
+        const fieldB = b.label.toLowerCase();
```
sjd78, line 261: You're not using the `sortField`.  I see the label being set in the switch above based on the grouping, but that leads to a different question.

Why have sorting on file name when the group by is incident and why have sorting on incident when the group by is file?

Sorting would be much more natural if the webview was an actual table instead of a collection of cards.

sjd78 (changes requested):
So my base question is: Why do we need sorting on multiple fields when the sort by fields are the same as the group by fields?  Just allow ascending/descending selection and always sort on the "label".

webview-ui/src/components/ViolationIncidentsList.tsx
djzager: Moved to only ascending vs descending

djzager:
https://github.com/user-attachments/assets/2c843496-ae1d-4c7a-9f23-b3144952cb38

sjd78 (approved):
LGTM!

PR #368, opened 2025-02-12
:ghost: vite deps and configs

sjd78:
Summary of work:
  - push the vite dependencies to the root `project.json` to share between all workspaces
  - use the `vite-plugin-dts` plugin to generate the type files for the shared project without having to also run `tsc` apart from `vite`
  - use the `vite-plugin-checker` plugin to check types within the builds and without having to also run `tsc` apart from `vite`

----

Note: This will also solve the dev mode problem where compiles and type checking of the vscode and webview-ui packages happen before the index.d.ts file of the shared package is generated.  Now, the shared project will fully compile with the index.mjs output being the last file generated that unblock the `wait-on` guards.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,6 +38,7 @@
         "@types/react-dom": "^18.3.0",
         "@typescript-eslint/eslint-plugin": "^8.7.0",
         "@typescript-eslint/parser": "^8.7.0",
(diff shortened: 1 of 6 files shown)


sjd78:
Closing since #374 includes these changes.

Issue #352, opened 2025-02-06
Resolution view: Support rendering different kinds of messages from kai-rpc-server


sjd78:
While a solution is running, the `kai-rpc-server` can send messages (progress updates etc) to be displayed to the user.

Add support to display at least 3 kinds of messages:
  - `SimpleChatMessage` --> Displayed as a standard string
  - `MarkdownMessage` --> The message content is markdown, nicely display it
  - `JsonMessage` --> The message is a JSON object, pretty print it as the message

Review the way all of the message are displayed.  A label component is probably not the correct component to hold the content we want to render.

References: 
  - https://github.com/konveyor/editor-extensions/pull/303#pullrequestreview-2596969038
  - https://www.patternfly.org/patternfly-ai/conversation-design
  - https://www.patternfly.org/patternfly-ai/chatbot/overview

**Note**: The view is chatbot like, but its detail messages require a bit more space to render, and the interactivity with the user is much more prescribed.

JonahSussman:
https://github.com/konveyor/kai/pull/652 added support for this on the kai binary

PR #372, opened 2025-02-14
:sparkles: Add support for markdown and json solution progress messages

sjd78:
Resolves: #352
Supports: https://github.com/konveyor/kai/pull/652
Built on: #374

Add `MarkdownChatMessage` and `JsonChatMessage` rendering to the resolution webview.

Markdown is configured to support:
  - Basic Markdown [1]
  - GitHub Flavored Markdown [2], [3]
  - HTML embedded in Markdown [4]
  - Syntax highlighting [5] in code \`\`\` blocks similar to what GitHub supports
  - Final string renders are sanitized so only "valid" HTML will make it to the page while invalid HTML is escaped and will show as normal text [6]

Json messages are formatted, wrapped in a Markdown code block, and syntax highlighting is used to render it as a message.

[1]: https://remarkjs.github.io/react-markdown
[2]: https://github.github.com/gfm
[3]: https://github.com/remarkjs/react-markdown
[4]: https://github.com/rehypejs/rehype-raw
[5]: https://github.com/rehypejs/rehype-highlight
[6]: https://github.com/rehypejs/rehype-sanitize

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6042,6 +6042,21 @@
         "url": "https://opencollective.com/unified"
       }
     },
+    "node_modules/hast-util-sanitize": {
+      "version": "5.0.2",
+      "resolved": "https://registry.npmjs.org/hast-util-sanitize/-/hast-util-sanitize-5.0.2.tgz",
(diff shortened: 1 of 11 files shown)


djzager (approved):
ACK

Issue #247, opened 2025-01-23
Kai Results page looks a little messy


midays:
some elements in the UI can be imporoved 

![Image](https://github.com/user-attachments/assets/15ce1790-a4a6-4b52-a095-d64225b5c938)

- number 1 and number 5 : this black padding can be removed by expanding the results panel
- number 2 this space is unnecessary the issue details can be expanded to the left
- number 3 what is the use for this text ?
- number 4 those look like a clickable button, maybe we can put a log panel that shows the high level output or include it in the existing output panel.

PR #373, opened 2025-02-14
:sparkles: Use @patternfly/chatbot library for the resolutions view

ibolton336:
Resolves https://github.com/konveyor/editor-extensions/issues/247

https://github.com/user-attachments/assets/adf894a4-0767-4257-bdf6-6d9da95196d5


rszwajko:
@ibolton336 @sjd78 
I don't have time to properly review the PR but let me share some thoughts:
1. problems with rendering custom components were one of the blockers to adopt the PF chat library. With that problem solved we are one step closer to adopt that solution.
2. we don't receive the timestamps from the server (or store them) but at this stage we could assume that it's just the time of displaying the message in the chat
3. this would need to be integrated with https://github.com/konveyor/editor-extensions/pull/372
4. it's not strictly related to this PR but it would be nice to get rid of "magic" flags in the ResolutionPage that simulate user-bot chat. We could keep a single list of messages in the state  and differentiate bot from user based on some propi.e. by `author`
```
export interface ChatMessage {
  author: "bot" | "user";
  kind: ChatMessageType;
  value: { message: string } | Record<string, unknown>;
  chatToken?: string;
  messageToken: string;
}
```
5. I like the Konveyor icon in the chat:)

djzager (dismissed):
VISACK. I am a +1 on this move.

sjd78:
Screencast looks good.  I tried running locally and couldn't get npm install to work.  Might have grabbed it before an update.

@ibolton336  can you rebase now that David's buttons are merged?

webview-ui/src/index.tsx
```diff
@@ -1,4 +1,9 @@
 import React from "react";
+import "@patternfly/patternfly/patternfly.css";
```
sjd78, line 2: The css is already imported below the component imports. I don't think it really matters, just should only import everything once.

sjd78 (changes requested):

ibolton336:
PF team feedback: 


> We've built most of the styling out assuming everything is built within a ChatBot, ChatbotContent, and MessageBox at a minimum. 

> Icons also you may want to inline or save locally.

> You're going to want to add stable timestamps in. 
> What Composer did was this: https://github.com/redhat-composer-ai/chatbot-ui/blob/main/src/app/BaseChatbot/BaseChatbot.tsx#L242
> There's a stored Messages array in state, and they apply timestamps before they resave that.

webview-ui/src/components/ResolutionsPage/ResolutionsPage.tsx
```diff
@@ -60,94 +49,69 @@ const ResolutionPage: React.FC = () => {
...
-            )}
-
-            {isResolved && !isFetchingSolution && (
-              <ReceivedMessage>All resolutions have been applied.</ReceivedMessage>
-            )}
-          </Flex>
-        </Flex>
+              )}
+              {hasEmptyResponse && !hasResponseWithErrors && (
+                <ReceivedMessage>Received response contains no resolutions.</ReceivedMessage>
```
sjd78, line 95: What is the difference between the text being `children` vs `mainContent`?  I'd prefer using one or the other to be consistent if there is no real difference.

webview-ui/package.json
```diff
@@ -18,12 +18,14 @@
...
   },
   "dependencies": {
+    "@patternfly/chatbot": "2.2.0-prerelease.29",
     "@patternfly/patternfly": "6.0.0-prerelease.15",
     "@patternfly/react-code-editor": "^6.0.0",
     "@patternfly/react-core": "6.0.0-prerelease.15",
     "@patternfly/react-icons": "^5.4.0",
     "@patternfly/react-table": "^5.4.1",
     "github-markdown-css": "^5.8.1",
+    "dompurify": "^3.2.4",
```
sjd78, line 28: Needed?  I don't see this being used anywhere.

shared/src/types/types.ts
```diff
@@ -113,6 +113,7 @@ export interface ChatMessage {
   value: { message: string } | Record<string, unknown>;
   chatToken?: string;
   messageToken: string;
+  timestamp: string;
```
sjd78, line 116: This annoys me for no apparent reason.  I know it's fine, but geeze Typescript, maybe make it a bit easier to say "ISO-8601 formatted datetime"?

sjd78 (commented):
Code looks solid.

I'm getting a strange layout in dev mode:
image

I'll try it in build mode

webview-ui/src/components/ResolutionsPage/ResolutionsPage.tsx
```diff
@@ -60,94 +48,69 @@ const ResolutionPage: React.FC = () => {
...
 
   return (
-    <Page
-      sidebar={
-        <PageSidebar isSidebarOpen={false}>
-          <PageSidebarBody />
-        </PageSidebar>
-      }
-    >
+    <Page>
```
sjd78, line 51:
```suggestion
    <Page isManagedSidebar={false} sidebar={<PageSidebar isSidebarOpen={false} />}>
```

webview-ui/src/components/ResolutionsPage/ResolutionsPage.tsx
ibolton336: Updated the props to just use "content" for string content and "extraContent" for React elements.

vscode/src/KonveyorGUIWebviewViewProvider.ts
```diff
@@ -165,7 +165,7 @@ export class KonveyorGUIWebviewViewProvider implements WebviewViewProvider {
       `style-src ${webview.cspSource};`,
       `font-src ${webview.cspSource};`,
       `connect-src ${webview.cspSource};`,
-      `img-src data: ${webview.cspSource};`,
+      `img-src data: ${webview.cspSource} http://${localServerUrl};`,
```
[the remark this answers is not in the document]
ibolton336: No it isn't - will remove.

webview-ui/src/index.css
```diff
@@ -45,3 +45,23 @@
   background-color: rgba(255, 255, 255, 0.5);
   color: brown;
 }
+
```
sjd78, line 48: Was tinkering with the layout, and if we do this:
```css
.pf-chatbot {
  height: 100vh;
}

.pf-chatbot__messagebox {
  width: 100%;
  max-width: 100% !important; /* to ignore media queries */
}

.pf-chatbot__jump {
  inset-inline-start: 95% !important;
}
```
then the chat view takes up more horizontal space with the scroll bar aligning with where the scroll bar is on the analysis view.


image

webview-ui/src/index.css
```diff
@@ -45,3 +45,23 @@
...
+}
+.pf-chatbot__message-text > details > summary {
+  cursor: pointer;
+}
+.pf-chatbot {
+  background-color: var(--vscode-editor-background) !important;
+}
+.pf-chatbot--embedded {
+  border: 1px solid var(--vscode-editor-foreground) !important;
+  border-radius: 20px;
```
sjd78, line 60: I'm not quite sold on the border here.  Turning it off looks ok (see the screenshot for the width changes).

If we want to keep it, I suggest using the color `--vscode-textSeparator-foreground` instead of the `--vscode-editor-foreground` so it isn't so bright/prominent.

sjd78 (changes requested):
A few layout tweaks

webview-ui/src/components/ResolutionsPage/bot_avatar.svg
sjd78, line 1: This one looks better on page: https://github.com/konveyor/community/blob/main/brand/logo/konveyor-logo-konveyor.svg

webview-ui/src/components/ResolutionsPage/ReceivedMessage.tsx
```diff
@@ -1,15 +1,49 @@
 import React from "react";
-import { FlexItem, Label } from "@patternfly/react-core";
+import { Message } from "@patternfly/chatbot";
+import rehypeRaw from "rehype-raw";
+import rehypeSanitize from "rehype-sanitize";
+import botAv from "./bot_avatar.svg";
```
sjd78, line 5:
```suggestion
import botAv from "./bot_avatar.svg?inline";
```

Force vite to inline the svg and resolve any headaches where the extension can't load the image. https://vite.dev/guide/assets#explicit-inline-handling

sjd78 (approved):
Looks really good!


https://github.com/user-attachments/assets/a632b692-8839-43e4-b796-d8cc67a2f4ad

PR #374, opened 2025-02-14
:seedling: Enable extension debugging, refactor the builds

sjd78:
With this change, running the "Run Extension" launcher should show breakpoints as active and hitting on the correct lines.  

:-)

Summary:
  - Align all of the `tsconfig.json` files around the root and only changing a minimum of setting per workspace.

  - Update the `launch.json` and `tasks.json` to be aware of source maps and locations.

  - Make sure source maps are generated properly for each workspace.

  - Push the vite dependencies to the root `project.json` to share between all workspaces.

  - Use the `vite-plugin-dts` plugin to generate the type files for the shared project without having to also run `tsc` apart from `vite`.

  - Use the `vite-plugin-checker` plugin to check types within the builds and without having to also run `tsc` apart from `vite`.

  - Update the webview-ui vite server to:
    - send a CORS header so dev mode viewing still works
    - use a base of `/out/webview` to match where the files will load from in a prod build

  - Update `KonveyorGUIWebviewViewProvider` CPS settings so all the file types can be loaded in dev and prod modes

  - Keep the optional dep on the rollup native package so the optional deps keep working for all platform/arch combos.


djzager (approved):
VISACK. This looks to me a drastic simplification. Thank you for doing this.

ibolton336:
> Use the vite-plugin-dts plugin to generate the type files for the shared project without having to also run tsc apart from vite.

> Use the vite-plugin-checker plugin to check types within the builds and without having to also run tsc apart from vite.

So awesome. Thank you, Scott.

Issue #379, opened 2025-02-18
[BUG] Can not apply changes to files that already have resolutions applied


shawn-hurley:
Steps to reproduce:

1. Fix a single incident in the `pom.xml` and then apply that change
2. Fix the rest of the incidents in pom.xml and then apply the change.

You will see that the outcome is a error notification about not being able to apply the diff, but also that then the merge view tries to get you to overwrite the file with just the patch file. 

This does seem to be solved if you commit the change in 1 before running 2, which makes me think the file is not reading read correctly when attempting to apply the diff.

rszwajko:
@shawn-hurley @djzager 
At first look it seems a known issue - please check [my response](https://github.com/konveyor/editor-extensions/issues/336#issuecomment-2631568775) in #336.
However please check if you can apply the patch manually via command line.

shawn-hurley:
closing this issue as it is now traced in https://github.com/konveyor/kai/issues/668

Issue #336, opened 2025-02-03
Reload last resolution adds diff whole output to the file


abrugaro:
When clicking on "Realod last resolution" button, an error is shown and, the file, displays the whole diff output with the number of additions and deletions

![Image](https://github.com/user-attachments/assets/b1529742-ce5e-4c51-8887-6327009e340e)

![Image](https://github.com/user-attachments/assets/4cf75d7e-619d-4d02-905f-7ef7e5d0388a)

![Image](https://github.com/user-attachments/assets/03c9a646-4f85-4de0-b859-b51b9117ec7a)

rszwajko:
@abrugaro 
This fallback is used when IDE is not able to apply the patch to the current source.
There might be various causes i.e. changes already applied, different branch checked out, manual edits since the time solution was requested etc

The limitation is caused by the requirement of being SCM-agnostic. If we would assume git-only then we could use commit hashes to detect and try to solve those cases. In the current code we could try to be more explicit with the error messages.

See also #332

abrugaro:
Thanks for the explanation @rszwajko I get it now!

> In the current code we could try to be more explicit with the error messages.

Even when displaying an error, the resolution still loads with the diff output.
Would it be possible to prevent this and show only the error?

rszwajko:
@abrugaro 
> Would it be possible to prevent this and show only the error?
>
Yes but then we should communicate it somehow that resolutions for some/all files are missing i.e. by marking those files as disabled or some error state.

abrugaro:
Makes totally sense.

I think it would be enough to just display an error like "the previous resolution can't be reloaded because the file was modified" 
Something like that and avoid loading the broken file

PR #380, opened 2025-02-18
:bug: Apply diff against modified file version

ibolton336:
Resolves https://github.com/konveyor/editor-extensions/issues/379

By unifying the read logic (i.e., reading from memFS if it exists, then falling back to the original workspace file when it doesn’t), we ensure we always have the correct “most up-to-date” file content before applying a new patch.

### DIFF
--- a/vscode/src/data/virtualStorage.ts
+++ b/vscode/src/data/virtualStorage.ts
@@ -4,7 +4,7 @@ import {
   Solution,
   SolutionResponse,
 } from "@editor-extensions/shared";
-import { Uri, window, workspace } from "vscode";
+import { FileSystemError, Uri, window, workspace } from "vscode";
 import { ExtensionState } from "src/extensionState";
 import * as Diff from "diff";
 import path from "path";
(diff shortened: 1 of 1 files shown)


rszwajko:
@ibolton336 
Some thoughts on the bug:
1. IMHO we should always apply the diff to the current files on the disk as this is our source-of-truth.
2. it seems that files in the workspace got out-of-sync with the files for which resolutions were requested - there is a simple way to test it - as I've mentioned in https://github.com/konveyor/editor-extensions/issues/379#issuecomment-2666663853 you need to take the diff and manually apply via command line. If it works then we have a bug.
3. `memFs` might contain a different source then the actual file i.e. using merge editor we merge read-only-source (the current source from the disk but made read only) with memFs-source (the same content as read-only but with the diff applied) and the result goes into the actual file (after saving the editor). During the merge step user can pick elements from both sides (read-only or memFs) AND manually edit the code. 
4. it might be easier to start with the original bug submitted by @abrugaro  in #336 and then extend to scenarios pointed by @shawn-hurley  in #379 and @sshaaf  in #383  which are more complex.

Issue #353, opened 2025-02-06
Provide a means from IDE for a user to select the "Effort" they want to expand on a given solution


jwmatthews:
As a user from the IDE I would like the ability to decide how much effort to expand on a given solution.
For example, perhaps I'm looking to resolve the javax -> jakarta incidents, I would like to chose to fix these and run as a "Low" effort.

Then later I select to fix JMS -> Smallrye, for this I would like to run as a "High" effort.

Or... maybe I run JMS->Smallrye as "Low", I don't like the results, I cancel them, then retry as "High".

The idea is as we are exploring different behaviors/results from the backend it would help to have controls in the IDE to make this exploration easier.  

To be clear, I'm looking for a UI driven means to quickly make this selection of Effort for all fixes we request, this way its easier to explore functionality from backend.   

Trying to avoid telling folks to manually update settings.json with ["konveyor.kai.getSolutionMaxEffort"](https://github.com/konveyor/editor-extensions/blob/main/vscode/package.json#L557-L563) for each time they want to change effort.

sjd78:
Related to #326

djzager:
Fixed with #381

PR #381, opened 2025-02-18
:sparkles: support selectable effort to getSolution

### LINKED ISSUE #326: Review the solution request actions on the Analysis view
If we can provide multiple potential solution request actions at any point in the Analysis view, the actions should be kebab based.  After #317 there is only a single action available at each scope.

----
              I prefer the kebab for the actions, but we can change the up later.

_Originally posted by @sjd78 in https://github.com/konveyor/editor-extensions/pull/317#discussion_r1936204730_

djzager:
* Puts the configured effort onto the ExtensionState and attempts to update it whenever the config is modified.
* Add effort types + helpers to shared.
* Introduce new component GetSolutionDropdown and use that wherever we want to render the action.

Fixes #326, #353


djzager:
Hey look it can be disabled

Uploading isDisabled.mov…

djzager:
It renders differently based on the provided scope:

Uploading Scoped.mov…

djzager:
Uploading ConfiguredEffort.mov…

It runs with configured effort

djzager:
Effort can be overridden

Uploading OverrideEffort.mov…

djzager:
Responds to config changes:

https://github.com/user-attachments/assets/6bfcd30e-6e54-4ad3-b7cb-99cc2ff6e9b6

rszwajko:
@djzager 
Functionality is great, but the UX needs some love. The buttons got really big. Visually they dominated the screen.
I liked the kebab from the original wireframes better. The split button does offer some advantage over kebab in terms of readability but the label/message needs to be shorter - ideally one word similar to well known actions like "Cancel".
We could consider adding extra confirmation modal with a brief summary of what we are triggering (at the cost of extra click).

djzager:
image

ibolton336:
Screenshot 2025-02-20 at 9 55 27 AM

Dropping 'Ascending / Descending' text will free up some space so we don't see the button wrap.

webview-ui/src/components/GetSolutionDropdown.tsx
```diff
@@ -0,0 +1,112 @@
...
+        isDisabled={isButtonDisabled}
+        splitButtonOptions={{
+          items: [
+            <MenuToggleAction
+              id="split-button-action-primary-example-with-toggle-button"
+              key="split-action-primary"
+              onClick={() => onGetSolution(incidents, state.solutionEffort)}
+              aria-label="Get solution"
+            >
+              Kai <Badge isRead>{incidents.length}</Badge>
```
ibolton336, line 57: Thoughts on moving to just the wrench Icon vs using "Kai" action verb?

webview-ui/src/components/GetSolutionDropdown.tsx
```diff
@@ -0,0 +1,112 @@
...
+          ],
+          variant: "action",
+        }}
+        onClick={() => setIsOpen(!isOpen)}
+        isExpanded={isOpen}
+        aria-label="Effort Levels"
+      />
+    ) : (
+      <MenuToggle
+        aria-label="kebab dropdown toggle"
```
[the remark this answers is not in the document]
ibolton336: An icon button to left to the kebab would also be more aligned with what I've seen done across other products. A "quick action" to fire a solution at the currently configured effort level. @JustinXHale Thoughts?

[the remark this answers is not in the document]
ibolton336: Screenshot 2025-02-20 at 10 23 26 AM

Or if we wanted to show the effort level in the "quick fix" button, might be cool to add a label.

ibolton336:
> Member

Tested the component & functionally it works awesome - The config changes work flawlessly for me. Just a few things to hash out with the UX.

djzager:
Make the button as small as I can so we can use it everywhere and be as least terrible as possible.

Uploading Screen Recording 2025-02-20 at 13.10.44.mov…

ibolton336:
Screenshot 2025-02-20 at 6 37 40 PM

Few things: 
- looks like we lost the disabled state when the server is not running
- the spacing of the button is really close to the line # / location column
- The button is now showing up in the incident table when it is in "view only" mode. see screenshot above. This is what the conditional code in incident table group was used for.

webview-ui/src/hooks/useExtensionState.ts
```diff
@@ -17,6 +17,7 @@ const defaultState: ExtensionData = {
   workspaceRoot: "/",
   chatMessages: [],
   solutionState: "none",
+  solutionEffort: "Low",
```
ibolton336, line 20: Could we pull in config value here?

webview-ui/src/components/IncidentTable/IncidentTableGroup.tsx
```diff
@@ -4,13 +4,9 @@ import { IncidentTable } from "./IncidentTable";
 
 export const IncidentTableGroup = ({
   onIncidentSelect,
-  onGetSolution,
-  workspaceRoot,
   incidents,
```
ibolton336, line 7: May just need an isReadOnly prop here to determine if the get  solution button is rendered conditionally.

webview-ui/src/hooks/useExtensionState.ts
djzager: I would think this would cross a boundary we don't want to...right? Calling `vscode/.../utilities` from the webview-ui is what I thought I was avoiding.

Maybe I'm missing something and there is a better way to do this than defaulting to "Low" knowing it'll be overwritten when we actually activate the extension (from extension.ts):

```
    this.data = produce(
      {
        ...
        solutionEffort: getConfigSolutionMaxEffortLevel(),
      },
```

ibolton336: Oh yeah that makes sense. Maybe we could just set it as undefined here then and only set it once we have the value.

webview-ui/src/components/GetSolutionDropdown.tsx
```diff
@@ -0,0 +1,101 @@
...
+  };
+
+  // State to track button disabled status
+  const [isButtonDisabled, setIsButtonDisabled] = useState(true);
+
+  useEffect(() => {
+    setIsButtonDisabled(
+      isFetchingSolution || isAnalyzing || isStartingServer || isInitializingServer,
+    );
+  }, [isFetchingSolution, isAnalyzing, isStartingServer, isInitializingServer]);
```
sjd78, line 41: `useMemo` is better here -- or even none since computing the 4 flags isn't compute intensive.

Also useEffect has a potential for a render blink while straight compute and useMemo don't.

```tsx
  const isButtonDisabled = useMemo(() => {
    return isFetchingSolution || isAnalyzing || isStartingServer || isInitializingServer;
  }, [isFetchingSolution, isAnalyzing, isStartingServer, isInitializingServer]);

```

sjd78 (commented):
I think you landed on a decent look of the button.  The text of the action menu items may need adjustment to indicate you're not changing the setting but rather choosing the setting for this action.

The other comments should be pretty straightforward

webview-ui/src/components/GetSolutionDropdown.tsx
djzager: My understanding was that with the useExtensionStateContext...I shouldn't need to do this anymore. But I should read more into useMemo versus useEffect.

webview-ui/src/components/IncidentTable/IncidentTable.tsx
```diff
@@ -43,31 +43,7 @@ export const IncidentTable: FC<IncidentTableProps> = ({
...
-                      {incidents.length === 1
-                        ? "Resolve 1 incident"
-                        : `Resolve ${incidents.length} incidents`}
-                    </Button>
-                  ),
-                }
-              : undefined
-          }
-        >
+        <CardHeader>
```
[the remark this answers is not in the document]
djzager: I had dropped it to reduce clutter and I was getting the feeling that no one who knew what they were doing were clicking this button. I don't feel very strongly.

image

webview-ui/src/components/GetSolutionDropdown.tsx
```diff
@@ -0,0 +1,94 @@
...
+    dispatch(getSolution(incidents, effort));
+  };
+
+  const isButtonDisabled =
+    state.isFetchingSolution || state.isAnalyzing || state.serverState !== "running";
+
+  const menuToggle = (
+    <MenuToggle
+      variant="primary"
+      size="sm"
```
ibolton336, line 39: Adding this causes the button to fall out of vertical alignment with respect to other toolbar items. Remove or add custom styling to center the button vertically.

djzager: #397 Created because the big button looks comical in the table.

webview-ui/src/components/GetSolutionDropdown.tsx
sjd78: For effect think "I need to go fetch something and then push data into my component" and memo is "I only want to do this calculation when the input values change".

webview-ui/src/components/IncidentTable/IncidentTable.tsx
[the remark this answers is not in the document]
sjd78: Yep that's the view.

Since we want to be able to solve any of the incidents, or group of incidents, or full issue, or all issues in a file, we do need to put actions just about everywhere.  I prefer the action on the message grouping so it is still a valid option.

in future PRs, we may be able to reasonably account for the "ah it's too much" factor with different indentation levels (which was my initial thought) and some different action wording.  What you have built is pretty solid.  We can tinker with the exact phrasing, location and render as needed if it still "feels off".

sjd78 (approved):
UX may need tweaking in the future, but it's decent for now.

Everything else looks really good.

ibolton336:
Nice work, David! Looking great.

PR #391, opened 2025-02-20
:bug: At extension startup, reload analysis result but do not reload solutions

ibolton336:
Resolves https://github.com/konveyor/editor-extensions/issues/336

- Remove the default solution load behavior on startup
- Remove the reload last solution button from the diff view
- Update to display the diff view when you hard load a solution from a file using the "Load Results from File" command. 
Screenshot 2025-02-20 at 3 45 18 PM

### DIFF
--- a/vscode/src/data/loadStaticResults.ts
+++ b/vscode/src/data/loadStaticResults.ts
@@ -50,11 +50,8 @@ const filePathsCorrect = (ruleSets: RuleSet[]) =>
     );
 
 export const loadResultsFromDataFolder = async () => {
-  const [analysisResults, solution] = await loadStateFromDataFolder();
+  const [analysisResults] = await loadStateFromDataFolder();
   if (analysisResults) {
     vscode.commands.executeCommand("konveyor.loadRuleSets", analysisResults);
-  }
-  if (solution) {
-    vscode.commands.executeCommand("konveyor.loadSolution", solution);
   }
 };
--- a/vscode/src/data/loadStaticResults.ts
+++ b/vscode/src/data/loadStaticResults.ts
@@ -33,6 +33,7 @@ export const loadStaticResults = async () => {
     }
   }
   if (solution) {
+    vscode.commands.executeCommand("konveyor.diffView.focus");
     vscode.commands.executeCommand("konveyor.loadSolution", solution);
     vscode.window.showInformationMessage("Successfully loaded the solutions");
   }
--- a/vscode/src/data/virtualStorage.ts
+++ b/vscode/src/data/virtualStorage.ts
@@ -75,8 +75,7 @@ export const writeSolutionsToMemFs = async (
     memFs.createDirectoriesIfNeeded(modifiedUri, KONVEYOR_SCHEME),
   );
 
-  const writeDiff = async (change: LocalChange) => {
-    const { originalUri, modifiedUri, diff, state } = change;
+  const writeDiff = async ({ diff, originalUri, modifiedUri }: LocalChange) => {
     const content = await applyDiff(originalUri, diff);
     memFs.writeFile(modifiedUri, Buffer.from(content), {
       create: true,
--- a/vscode/src/data/virtualStorage.ts
+++ b/vscode/src/data/virtualStorage.ts
@@ -75,12 +75,19 @@ export const writeSolutionsToMemFs = async (
     memFs.createDirectoriesIfNeeded(modifiedUri, KONVEYOR_SCHEME),
   );
 
-  const writeDiff = async ({ diff, originalUri, modifiedUri }: LocalChange) => {
-    const content = await applyDiff(originalUri, diff);
-    memFs.writeFile(modifiedUri, Buffer.from(content), {
-      create: true,
-      overwrite: true,
-    });
+  const writeDiff = async (change: LocalChange) => {
+    const { originalUri, modifiedUri, diff, state } = change;
+    if (state === "applied") {
+      const msg = `Already applied solution for ${originalUri.path}.`;
+      window.showErrorMessage(msg);
+      return;
+    } else {
+      const content = await applyDiff(originalUri, diff);
+      memFs.writeFile(modifiedUri, Buffer.from(content), {
+        create: true,
+        overwrite: true,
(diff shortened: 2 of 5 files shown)


rszwajko:
@ibolton336 
Re-loaded solutions need to be in `pending` state. The use case behind is the following:
1. receive the solutions (possibly after waiting a long time)
2. review the solutions - apply/ignore/merge
3. at some point user decides that he wants to start from scratch
4. discard the already applied changes in SCM (i.e. git)
5. re-load the solutions 
6. start reviewing once again

Effectively it's a solution history but with depth 1. In the wireframes we have the design for full history - check https://github.com/konveyor/editor-extensions/blob/main/docs/roadmap/wireframes/18.kai-resolutions-tree-view.excalidraw.svg

shawn-hurley:
I think then, at that time, they would want a feature for the history; until then, I think this current behavior of whatever was chosen in the past has to be re-chosen when I restart is not a good UX. I am a +1 on this change,

vscode/src/data/virtualStorage.ts
```diff
@@ -75,12 +75,19 @@ export const writeSolutionsToMemFs = async (
...
-    const content = await applyDiff(originalUri, diff);
-    memFs.writeFile(modifiedUri, Buffer.from(content), {
-      create: true,
-      overwrite: true,
-    });
+  const writeDiff = async (change: LocalChange) => {
+    const { originalUri, modifiedUri, diff, state } = change;
+    if (state === "applied") {
+      const msg = `Already applied solution for ${originalUri.path}.`;
+      window.showErrorMessage(msg);
```
shawn-hurley, line 82: I don't know if an error should be here, I think a log message, but if it is already applied then good things are happening not bad IMO.

djzager: I think `window.showInformationMessage` is a better choice.

https://code.visualstudio.com/api/references/vscode-api#window.showInformationMessage

djzager:
> Re-loaded solutions need to be in `pending` state.
> Effectively it's a solution history but with depth 1. In the wireframes we have the design for full history - check

I agree that a proper solution history would be valuable feature, and the wireframes illustrate a more comprehensive approach to handling past resolutions. However, the current implementation does not yet support full history tracking, so treating previously applied solutions as "pending" is causing problems. Since we do not maintain a structured historical record of solutions, we can’t reliably differentiate between solutions that were intentionally discarded vs. ones that should be reconsidered.

Because of that, I think this change makes sense for now. I’m +1 on this approach, but I also agree that we should discuss how to properly introduce solution history in a way that provides a better user experience in the future.

ibolton336:
> > Re-loaded solutions need to be in `pending` state.
> > Effectively it's a solution history but with depth 1. In the wireframes we have the design for full history - check
> 
> I agree that a proper solution history would be valuable feature, and the wireframes illustrate a more comprehensive approach to handling past resolutions. However, the current implementation does not yet support full history tracking, so treating previously applied solutions as "pending" is causing problems. Since we do not maintain a structured historical record of solutions, we can’t reliably differentiate between solutions that were intentionally discarded vs. ones that should be reconsidered.
> 
> Because of that, I think this change makes sense for now. I’m +1 on this approach, but I also agree that we should discuss how to properly introduce solution history in a way that provides a better user experience in the future.

I need to figure out a way to keep filesystem and memfs in sync... 

If I reload last solutions, and they have already been applied, then I am in a pickle because what if I reverted the last solutions in the filesystem... and now I want to re-apply solutions. The solutions are marked as applied & now I will need to cross check the filesystem to see if the filesystem has the solution diff applied.

djzager:
We may need to use the index from the diff itself in order to make a decision on that:

```
diff --git a/webview-ui/src/components/IncidentTable/IncidentTable.tsx b/webview-ui/src/components/IncidentTable/IncidentTable.tsx
index 9c2cf4e..1b48c27 100644
--- a/webview-ui/src/components/IncidentTable/IncidentTable.tsx
+++ b/webview-ui/src/components/IncidentTable/IncidentTable.tsx
@@ -37,11 +37,7 @@ export const IncidentTable: FC<IncidentTableProps> = ({ incidents, message, onIn
   return (
     <>
       <Card isPlain>
-        <CardHeader
-          actions={{
-            actions: <GetSolutionDropdown incidents={incidents} scope="in-between" />,
-          }}
-        >
+        <CardHeader>
           <Markdown>{message}</Markdown>
         </CardHeader>
```

index 9c2cf4e..1b48c27 100644
- This line shows the Git blob hashes of the file before and after the change.
- 100644 is a file mode indicator, meaning it’s a normal file (not executable, not a symlink).

djzager:
I believe this is the behavior we are trying to prevent.

https://github.com/user-attachments/assets/42d8be33-6d5c-4509-91c3-0b9b84e4a24b

ibolton336:
https://github.com/user-attachments/assets/cb0b7f74-8cf1-4f9d-bc6e-995e1caf1762

shawn-hurley:
@ibolton336 I think that in this particular case, we wait for a user to yell. I assume I can edit the solution to move from applied to pending or make this go if I have to, too?

shawn-hurley (approved):

djzager (approved):

vscode/src/data/loadStaticResults.ts
```diff
@@ -33,6 +33,7 @@ export const loadStaticResults = async () => {
     }
   }
   if (solution) {
+    vscode.commands.executeCommand("konveyor.diffView.focus");
```
sjd78, line 36: Oddly, this command doesn't exit.

djzager:
> The diff showing the commit id is not part of the user's actual repo. We have to be scm agnostic. If the user is using mercurial or bitbucket, or visual source safe (for some unknown reason), those commit ids really don't mean anything.
> 
> I believe those commit ids are actually from the "vfs git" that kai builds to track changes. Once the solution is sent back, those commits should cease to exist.

Good point.

sjd78 (approved):
Users will yell if they start reviewing a solution, but then close down vscode, and come back later.  The manual reload looks like it is still an option in that scenario.

PR #396, opened 2025-02-25
:bug: Context API for extension state management

ibolton336:
- Prevent prop drilling
- Ensure there's only one source of truth for the state
- Avoid unnecessary re-renders from multiple state hooks
- Make the state management more maintainable

The context is set up at the root level, so any component in the app tree can access the state without needing to pass it through props.


sjd78 (approved):
LGTM -- nice switch to share extension level state

PR #399, opened 2025-02-26
:bug: respect isReadOnly for solution dropdown

### LINKED ISSUE #398: GetSolution button should not be rendered in incident table when in  "read-only" mode
Image

### LINKED ISSUE #400: Issue table styling issues
-  change that button to a link so we can avoid that little indentation at the beginning of the “shoppingCartOrderProcessor.java” field
-  make sure the col widths are set / matching between the header cols and row cols. Shouldnt need to truncate location title if we don’t have to. Also we need some space between the row data. 

Image

djzager:
This PR fixes problems with rendering the incident table(s).
* When isReadOnly, we shouldn't be rendering the solution dropdown
* When isReadOnly, we don't actually need to full rendered path because it gets cluttered
* We need to handle windows paths with care, not just for relative paths but when computing the filename from the incident.uri
* Some other minor tweaks to the look and feel of the table and solution dropdown
 
Fixes #398
Fixes #400

### DIFF
--- a/webview-ui/src/components/IncidentTable/IncidentTable.tsx
+++ b/webview-ui/src/components/IncidentTable/IncidentTable.tsx
@@ -38,6 +38,7 @@ export const IncidentTable: FC<IncidentTableProps> = ({
   const ISSUE = "Issue";
   const LOCATION = "Location";
   const FOLDER = "Folder";
+
   return (
     <>
       <Card isPlain>
@@ -60,9 +61,8 @@ export const IncidentTable: FC<IncidentTableProps> = ({
                 <Tr>
                   {isReadOnly ? (
                     <>
-                      <Th width={15}>{ISSUE}</Th>
-                      <Th width={60}>{FOLDER}</Th>
-                      <Th width={25}>{LOCATION}</Th>
+                      <Th width={60}>{ISSUE}</Th>
+                      <Th width={40}>{LOCATION}</Th>
                     </>
                   ) : (
                     <>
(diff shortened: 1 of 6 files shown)


webview-ui/src/components/IncidentTable/IncidentTable.tsx
```diff
@@ -44,7 +44,13 @@ export const IncidentTable: FC<IncidentTableProps> = ({
     <>
       <Card isPlain>
         <CardHeader
-          actions={{ actions: <GetSolutionDropdown incidents={incidents} scope="in-between" /> }}
+          actions={
```
ibolton336, line 47: Cleaner option could be:

actions={!isReadOnly && {
  actions: <GetSolutionDropdown incidents={incidents} scope="in-between" />
}}

djzager: tried this. It specifically wanted `undefined` or the `{ actions: ... }` so I updated it at the bottom of the table.

webview-ui/src/components/IncidentTable/IncidentTable.tsx
```diff
@@ -54,19 +58,34 @@ export const IncidentTable: FC<IncidentTableProps> = ({
             <Table aria-label="Incidents" variant="compact">
               <Thead>
                 <Tr>
-                  <Th>{ISSUE}</Th>
-                  <Th width={50}>{FOLDER}</Th>
-                  <Th>{LOCATION}</Th>
-                  <Th />
+                  {isReadOnly ? (
+                    <React.Fragment>
```
sjd78, line 62: `<>` === `<React.Fragment>`

sjd78 (commented):
Looks reasonable.

djzager:
After this fix...

The resolution details.
image

Analysis page fix for files at the root of the workspace:
image

Additionally this should fix the paths when grouping by file on windows

djzager:
Updated styling:
image

image

image

sjd78 (approved):
Changes look good to me.  The link style button looks good.  No buttons in "read only" mode is working great.

sjd78 (approved):
I'm indifferent to dropping the folder on the resolution view of the incident list.

ibolton336 (approved):

Issue #207, opened 2025-01-17
Do not run analysis when you make changes under .vscode


pranavgaikwad:
There is no need to re-run analysis when files under "excluded" dirs change.

I think this may be related to effort behind https://github.com/konveyor/editor-extensions/issues/202.

Issue #202, opened 2025-01-16
[Feature] UX for ignored directories


djzager:
konveyor/kai#452 gives us the ability to tell kai (and the analyzer) to ignore specific directories during analysis. We should provide users with more flexibility to define which directories are ignored during analysis. This feature would empower users to customize their excluded paths while maintaining compatibility with the analyzer's requirements.

**Default Behavior:** Automatically exclude the .vscode directory, allowing users to add or remove exclusions as needed.

Considerations:
* Could we use `.gitignore` with something like https://www.npmjs.com/package/globby to just ignore everything git would ignore.
* Do we need to use a `.konveyorignore`?
* Maybe we could just expose this using .contributes.configuration in our vscode package.json

Other thoughts:
* .gitignore syntax is complex and not entirely compatible with the analyzer, especially for patterns like ! and **.
* For now, sending /path/to/workspace/.vscode directly to the analyzer is a practical stopgap measure to address immediate issues and enable the upcoming release.

This feature would enhance user control while ensuring the analyzer operates effectively with ignored paths.

PR #415, opened 2025-03-01
:sparkles: Add configurable ignores for analysis

sjd78:
Setup the use of ignore files for two things:

1. When saving a file, only send the file to partial analysis if it is not configured to be ignored.

2. When initializing the kai jsonrpc server, use the ignore files to find directories to pass to the analyzer as excluded paths.

Documentation has been added to vscode/README.md

Resolves: #202
Resolves: #207
Resolves: #405


vscode/src/client/analyzerClient.ts
```diff
@@ -694,7 +693,7 @@ export class AnalyzerClient {
     return true;
   }
 
-  public getAnalyzerPath(): string {
+  protected getAnalyzerPath(): string {
```
djzager, line 696: For my edification. Why `public` -> `protected`?

vscode/src/paths.ts
```diff
@@ -90,3 +94,96 @@ export function fsPaths(): ExtensionFsPaths {
   }
   return _fsPaths;
 }
+
+const DEFAULT_IGNORES = [".git", ".vscode", "target", "node_modules"];
+const IGNORE_FILE_IN_PRIORITY_ORDER = [".konveyorignore", ".gitignore"];
+
+let _ignoreByFunction: undefined | ((path: string) => boolean);
```
djzager, line 101: Again, for my edification. Is the order of the typing here significant?

vscode/src/paths.ts
```diff
@@ -90,3 +94,96 @@ export function fsPaths(): ExtensionFsPaths {
...
+
+/**
+ * Check a Uri to see if it should be ignored by a partial analysis on save.
+ */
+export const isUriIgnored = (uri: vscode.Uri): boolean => {
+  if (uri.scheme !== "file") {
+    return true;
+  }
+
+  const f = relative(fsPaths().workspaceRepo, uri.fsPath);
```
djzager, line 142: This looks like the method we should be using when getting relativeDir for incident uris.

djzager (commented):
VISACK. I'm going to pull and play with it to make sure.

vscode/src/client/analyzerClient.ts
sjd78: Just being pedantic.  `protected`  because it isn't intended to be called outside of the class (it's a helper function).  So all the public functions are ok to call from the other extension code.

vscode/src/paths.ts
sjd78: Not really.  The variables starts off as an undefined and then will become a function once the first call is made.  So the order here is just the order I was writing the code.

vscode/src/paths.ts
sjd78: Could be for sure.  It may need some extra work to go through the posix handling code to normalize input of path separators, and then to output with the platform correct path separators.  :thinking:

ibolton336:
Tested and it's working for me.

ibolton336 (approved):
Nicely done @sjd78. Will make testing even faster by narrowing down to a single dir as a side benefit.

PR #421, opened 2025-03-05
:bug: load the results even when no rulesets

### LINKED ISSUE #370: Analysis results show incidents even when there are no incidents
Image

After accepting the resolution, the analysis runs and shows that no incidents were found. but, the Analysis View tab still displays incidents.

djzager:
Fixes #370

### DIFF
--- a/vscode/src/client/analyzerClient.ts
+++ b/vscode/src/client/analyzerClient.ts
@@ -539,11 +539,9 @@ export class AnalyzerClient {
           }
           if (ruleSets.length === 0) {
             vscode.window.showInformationMessage("Analysis completed. No incidents were found.");
-            this.fireAnalysisStateChange(false);
-            return;
           }
 
-          vscode.commands.executeCommand("konveyor.loadRuleSets", ruleSets, filePaths);
+          vscode.commands.executeCommand("konveyor.loadRuleSets", ruleSets);
           progress.report({ message: "Results processed!" });
           vscode.window.showInformationMessage("Analysis completed successfully!");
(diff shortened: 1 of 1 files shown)


djzager:
> Does this still work in a partial analysis?

Don't know yet. Might need to distinguish somehow. But this is to try and see if it fixes specifically what @savitharaghunathan is seeing when we do a full analysis and the results are empty legitimately.

shawn-hurley (dismissed):
VISACK

savitharaghunathan:
tested this without my custom rule and the partial analysis works.

vscode/src/client/analyzerClient.ts
```diff
@@ -539,11 +539,9 @@ export class AnalyzerClient {
           }
           if (ruleSets.length === 0) {
             vscode.window.showInformationMessage("Analysis completed. No incidents were found.");
-            this.fireAnalysisStateChange(false);
-            return;
           }
 
-          vscode.commands.executeCommand("konveyor.loadRuleSets", ruleSets, filePaths);
+          vscode.commands.executeCommand("konveyor.loadRuleSets", ruleSets);
```
djzager, line 544: I did this because the loadRuleSets command only takes 1 argument (ruleSets).

djzager:
I will include the issue to konveyor/java-analyzer-bundle once one exists. We did verify that this fix does not negatively impact partial analysis results AND when the analyzer reports 0 incidents -- we report 0 incidents.

ibolton336 (approved):

Issue #425, opened 2025-03-06
Analysis view is not opened on initial load


ibolton336:
- There is no intuitive way to open the analysis page when a user is first setting up the IDE extension. We should display the analysis page by default when the konveyor kai sidebar menu item is clicked. 
- Currently, we have to find the book icon and click that from the side issues pane to open up the analysis page.

sjd78:
Also -- when the extension is reloaded, I've ended up collecting a number of analysis views open.

Closing any of the editor views when the extension is deactivated / reloaded would also be a good thing.

pranavgaikwad:
@sjd78 @ibolton336 The issue I am about to describe sounds very similar to this. When I click "Start Analyzer", it doesn't navigate me to the analysis view. Unsure if it's a bug... it'd be great if navigation worked.

https://github.com/user-attachments/assets/6aa4f3a1-4ca6-4e32-8fa2-c71b6ec289e9

jwmatthews:
Moving this off v0.1.0, we have a partial fix in #433, there is likely a bit more we want to handle but don't plan to have that in 0.1.0

pranavgaikwad:
@jwmatthews Is there anything specific you were looking for the fix?

PR #432, opened 2025-03-06
:bug: Show analysis panel when KAI extension initialized

ibolton336:
Resolves https://github.com/konveyor/editor-extensions/issues/425


rszwajko:
@ibolton336 
in the past this used to work but was a side effect of the Issue tree view
https://github.com/konveyor/editor-extensions/blob/1ef4856803893aea61ba91ee21d1ee04aa2d0fec/vscode/src/issueView/register.ts#L52

It would be nice to re-implement in a less hacky way.
In my opinion the view should open the first time we enter Konveyor perspective but feel free to disagree :)

PR #433, opened 2025-03-06
:bug: Redirect user to analysis page at the end of the konveyor walkthrough

ibolton336:
- Addresses the comment from 
[@Pranav](https://redhat-internal.slack.com/team/UJVF3G38W) on https://github.com/konveyor/editor-extensions/issues/425 - the walkthrough button isn’t stateful so we were able to start the server but there were no notifications or button state on the walkthrough that let us know that the server was starting/initializing. You could click it a bunch of times and kick off several instances of the server. The fix in PR 433 just ushers the user to the analysis page and encourages the user to start the server from there.

Screenshot 2025-03-07 at 2 31 35 PM

### DIFF
--- a/vscode/media/walkthroughs/open-analysis-panel.md
+++ b/vscode/media/walkthroughs/open-analysis-panel.md
@@ -1,6 +1,6 @@
-# Start Server
+# Open Analysis Panel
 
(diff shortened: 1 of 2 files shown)


djzager (approved):
This is actually a really great way to handle this right now.

Issue #334, opened 2025-01-31
collect-assets could use cache checking to skip downloading assets that haven't been updated


sjd78:
Specifically to help out developers, the `collect-assets.js` scripts could use basic caching HTTP headers to be able to skip downloading of any release assets (or workflow artifacts).  If an asset hasn't been updated since the last download, skip the download.

`ETag`/`If-None-Match` headers could be used.  The ETag would need to be stored and accessed in future (from our collect.json maybe?).

`Last-Modified`/`If-Modified-Since` request header on the asset fetch could be used.  The `If-Modified-Since` header could be generated from the mtime of the previously downloaded file.

In either case, download could always be forced by removing the download_assets folder and running the script again.

Issue #234, opened 2025-01-21
Update collect-assets to download static assets published by kai


sjd78:
Replace assets currently stored in the repo with downloads of those assets that have been published by kai to a release (>=0.0.5)  or by a workflow run's artifacts.  Target assets are:
  - seed rulesets
  - jdtls bundles
  - maven opensource labels file
  - jdtls runtime (currently downloaded directly from eclipse)

-----

> Note: #194 takes care of the first change.
> 
> Followup issues are required to 
>   1. replace static assets with downloading those assets packaged with kai (jdtls, jdtls bundles, seed rulesets, maven index file)
>   2. make the collection/download fail if the target release/workflow run is not available 

 _Originally posted by @sjd78 in [#198](https://github.com/konveyor/editor-extensions/issues/198#issuecomment-2605530435)_

PR #453, opened 2025-03-24
:seedling: Update collect-assets scripts

sjd78:
Resolves: #234
Resolves: #334

Update the collect-assets script such that assets can be sourced from GitHub release assess or from a GitHub workflow's artifacts.  Download of assets and artifacts is separated from unpacking of the assets.

When complete, the full set of downloaded assets that the editor-extensions can use will be located under `{project_root}/downloaded_assets`.

Downloaded files will have the ETag stored in a sidecar file.  This allows future runs to skip the download step if the url's data has not been changed.

Note: File globs are used during unpacking to allow extracting only a portion of an archive that is
relevant to the desired file set.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ out/
 build/
 dist/
 downloaded_assets/
+downloaded_cache/
 *.vsix
 
 # TypeScript build info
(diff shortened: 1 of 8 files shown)


djzager (approved):
Played around with this locally. lgtm.

PR #455, opened 2025-03-24
:seedling: drop repo asset `fernflower`

sjd78:
The fernflower java decompiler is not referenced or sent to the Kai analysis server at this time.

Drop the jar from the repo, but leave the path references in place.  If it is needed in the future, a download step can be added to `collect-assets.js`, and a copy step added to `copy-dist.js`.

Follows: #453
Partial: #234


djzager (approved):
This one is easy 😁

Issue #460, opened 2025-03-26
Usage of vscode context Keys to reflect current extension state


ibolton336:
# Rework Context Keys for Extension Commands

## Description
Tie vscode context keys into current event emitter logic 

## Goals
1. **Cleanly define context keys**: Minimal but sufficient to reflect relevant extension states (e.g., config availability, server running, analysis ready).
2. **Refactor usage**: Remove or rename outdated keys, and apply the newly standardized keys in `package.json` `when` clauses.

## Tasks

1. **Audit Existing Keys**  
   - Identify all current context keys (e.g. `konveyor.configAvailable`, `konveyor.analysisReady`, etc.).  
   - Note how/where they are used in `package.json` (`when` clauses), commands, or the walkthrough’s completion events.

2. **Define New or Updated Keys**  
   - Decide on the final set of context keys we need (e.g. `konveyor.serverRunning`, `konveyor.configReady`, etc.).  
   - Standardize naming (i.e. `konveyor.somethingCamelCase`).

3. **Refactor Extension Logic**  
   - Update the extension code to set or unset these keys as state changes (e.g. `vscode.commands.executeCommand('setContext', 'konveyor.serverRunning', true)` when the analyzer starts).  
   - Remove or rename any outdated or unused keys.  
   - Update `package.json` `when` clauses to rely on the new keys.

4. **Walkthrough Integration**  
   - Where relevant, swap existing completion events with context-based events, like `onContext:konveyor.serverRunning == true`.

## Acceptance Criteria
- [ ] **Context keys accurately reflect extension state** (e.g. server status, config status, analysis readiness).  
- [ ] **Commands show/hide predictably** based on the relevant keys (for example, “Start Server” only shows if `konveyor.serverRunning == false`).  
- [ ] **Walkthrough steps** that depend on context keys complete correctly when the key conditions are met.

PR #464, opened 2025-03-27
:bug: Only complete analysis arguments walkthrough step once valid labels selected

ibolton336:
Resolves part of https://github.com/konveyor/editor-extensions/issues/460 

https://github.com/user-attachments/assets/b68f9d4c-f095-4ee0-809f-42baf4b329fc


sjd78 (commented):
Given that other config related changes are in play, why not have a config watcher class/function/whatever that just registers listeners, provides a disposable (so all the listeners can be removed when the extension is stopped/started), and then just emits context changes and/or config change events to other components?

It could just be a fancy event emitter with a few handlers like "onConfigurationStatusChage" or "onAnalysisConfigChange" or similar.

Then an `onConfigurationStatusChange` handler could exist to set context variables that are referenced in package.json when clauses etc.

The `onAnalysisConfigChange` handler could force the analyzer to restart if it is running etc.

Similar handlers could be used to push config changes to the webviews when relevant config changes happen.

PR #475, opened 2025-04-09
:sparkles: Replace Kai binary with analyzer binary

pranavgaikwad:
What this does: 
- Replaces Kai binary in start() with analyzer binary
- Removes init() method as init happens in start
- Makes sure lifecycle of the analyzer binary is managed as expected

Outstanding work:
- Make analyze() and stop() functions actually work - this needs analyzer to work with JSON-RPC messages with framing


vscode/package.json
```diff
@@ -441,14 +441,7 @@
...
-            "TRACE",
-            "DEBUG",
-            "INFO",
-            "WARN",
-            "ERROR",
-            "CRITICAL"
-          ],
+          "pattern": "^(TRACE|DEBUG|INFO|WARN|ERROR|CRITICAL)(?:[+-]\\d+)?$",
           "default": "DEBUG",
           "description": "Logging level used for the server binaries",
```
djzager, line 446: Update the description?

shared/src/types/types.ts
```diff
@@ -140,7 +140,6 @@ export type ServerState =
   | "configurationNeeded"
   | "configurationReady"
   | "starting"
-  | "readyToInitialize"
   | "initializing"
```
djzager, line 143: Should probably be dropping this one two as it appears you are removing the change to this state.

vscode/src/client/analyzerClient.ts
```diff
@@ -18,30 +18,26 @@ import {
   SolutionState,
   Violation,
 } from "@editor-extensions/shared";
-import { paths, fsPaths, ignoresToExcludedPaths } from "../paths";
```
djzager, line 21: Are we losing this capability? Maybe I missed it but I do not see anything in starting the analyzer or running the analysis that handles this.

vscode/src/client/types.ts
```diff
@@ -37,33 +37,3 @@ export interface KaiConfigModels {
   llmRetries?: number;
   llmRetryDelay?: number;
 }
-
```
djzager, line 40: I don't know why, but I love code deletions.

djzager (commented):
I know it's still in draft but these are the things I see.

shared/src/types/types.ts
pranavgaikwad: Thanks, didn't fully know what was used where, was going to do a thorough review once basic changes are in. This helps.

vscode/src/client/analyzerClient.ts
pranavgaikwad: We are not, right now, analysis isn't working. I have this under outstanding work, but thanks for pointing out, I definitely would have missed this detail.

djzager:
A lot has changed since this was originally put up. Is this something you will be working to bring in?

PR #478, opened 2025-04-10
:seedling: there can be only one lint in ci checks

djzager:


sjd78 (changes requested):
Changes to the workflow LGTM.  Won't make it any faster but will reduce the number of annotations that land!

However, the linter is also called by the `pretest` script in `vscode/package.json` and the `test` script in `package.json`.  Would be best to remove the linter bit of those scripts.

`git grep "run lint"` to see all the places

sjd78 (approved):
LGTM

Now there are only 4 annotations that hit!
image

PR #481, opened 2025-04-17
:sparkles: Feature/hot rerun of analysis

### LINKED ISSUE #472: :sparkles: Trigger rerun of analysis when working copy is changed
Blocked by #471 , as we need to use VSCodes jdtls in order to run analysis on the working copy rather than what's on disk

Keep analysis diagnostic info up to date as filestate (in-editor) changes, by triggering a partial analysis

## Out of scope
- Support for anything but the Java provider calls, as there is no existing easy mechanism to feed the working copy to the other providers yet

pranavgaikwad:
Fixes #472 

Introduces hot rerun of analysis in that analysis is run whenever - 
1. A file is saved (immediately)
2. A file is modified (after notifying file changes to the backend with a backoff)
   - File changes are communicated to the backend, the backend is supposed to handle in-memory changes
Makes sure server is not overwhelmed at any point using backoffs but we may have to tweak the numbers as we test this more. So far its working well.

Needs https://github.com/konveyor/kai/pull/764

### DIFF
--- a/shared/src/types/types.ts
+++ b/shared/src/types/types.ts
@@ -132,6 +132,7 @@ export interface ExtensionData {
   isStartingServer: boolean;
   isInitializingServer: boolean;
   isContinueInstalled: boolean;
+  isAnalysisScheduled: boolean;
   serverState: ServerState;
   solutionState: SolutionState;
   solutionData?: Solution;
--- a/vscode/src/analysis/backoffManager.ts
+++ b/vscode/src/analysis/backoffManager.ts
@@ -1,0 +1,64 @@
+export class BackoffManager {
+  // next scheduled analysis runs after this delay
+  private backoffTimer: NodeJS.Timeout | null = null;
+  // reset the backoff delay when we are idle for a while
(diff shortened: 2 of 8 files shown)


ibolton336 (changes requested):
Looking good - Do need to define a default value 

https://github.com/konveyor/editor-extensions/blob/ba60375d382824509ea5badb936d58632e3875e0/vscode/src/extension.ts#L56

for the isAnalysisScheduled boolean here.

pranavgaikwad:
@ibolton336 thanks for catching that!

ibolton336 (approved):
LGTM - do we need to get rid of code around "analyzeOnSave" since this seems to bypass that setting?

pranavgaikwad:
@ibolton336 Okay I wasn't aware of that code. But I will create a new issue for myself to address that in a followup. Does that sound good?

ibolton336:
> @ibolton336 Okay I wasn't aware of that code. But I will create a new issue for myself to address that in a followup. Does that sound good?

Works for me - ty.

PR #491, opened 2025-05-01
:sparkles: Introduces a shared library for agentic workflows

pranavgaikwad:


agentic/package.json
```diff
@@ -0,0 +1,40 @@
...
+    },
+    "./package.json": "./package.json"
+  },
+  "scripts": {
+    "clean": "rimraf dist",
+    "ci": "npx jest",
+    "lint": "eslint .",
+    "lint:fix": "eslint --fix .",
+    "prebuild": "npm run clean",
+    "build": "npm run prebuild && tsup src/index.ts --dts --format esm --out-dir dist --clean",
```
sjd78, line 22: npm runs the prebuild script before the build script, so explicitly calling it should be redundant

agentic/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "dev": "tsup src/index.ts --dts --format esm --out-dir dist --watch"
```
sjd78, line 23: why not just reuse the vite config from shared instead of having another build tool?

vscode/src/commands.ts
```diff
@@ -26,6 +26,8 @@ import {
   ChatMessageType,
   GetSolutionResult,
 } from "@editor-extensions/shared";
+import { KaiWorkflowMessage, KaiWorkflowMessageType } from "../../agentic/src/types";
```
sjd78, line 29: You should import from the actual project name instead of directly referencing the source file.  That way you're using the built output as a library and not pulling in the code directly.

Should look like:
```suggestion
import { KaiWorkflowMessage, KaiWorkflowMessageType } from "@editor-extensions/agentic";
```

You'll need to rebuild the agentic workspace anytime you make changes so it can be used by the other projects (there is no auto library cross compile in typescript/tsc/javascript/npm land)

package.json
```diff
@@ -37,11 +37,12 @@
     "extra-types",
     "shared",
     "webview-ui",
+    "agentic",
```
sjd78, line 40: Be aware that order in the workspaces list here matters for any script that has `npm ... --workspaces` calls .  If agentic is referenced by webview-ui at all, it should come before it.

sjd78 (commented):
Just a few thoughts about the package setup.

agentic/package.json
pranavgaikwad: I am not very familiar with the build tooling, when i tried to use vite config, it didnt work because its building it for webviews which dont have filesystem apis needed by langgraph. So i chose the path of least resistance right now, but if you guys can help me fix those issues, happy to learn.

vscode/src/commands.ts
pranavgaikwad: Yeah, this only exists so that i can debug easily with the debugger. Going to make it come from dist once pr is ready for review

agentic/package.json
pranavgaikwad: @sjd78 can you help me with above?

ibolton336: I don't think anything in /agentic needs to be bundled for the browser, right?

djzager (dismissed):

ibolton336 (dismissed):

PR #510, opened 2025-05-15
:sparkles: authenticate with solution server

djzager:


vscode/src/commands.ts
```diff
@@ -84,6 +84,23 @@ const commandsMap: (state: ExtensionState) => {
         console.error("Could not start the server", e);
       }
     },
+    "konveyor.getToken": async (): Promise<string | undefined> => {
+      let token = await state.extensionContext.secrets.get("konveyor.token");
+
+      if (!token) {
+        // TODO(djzager): Update this to get the user's password
+        // once that logic is implemented in Hub/SolutionServer
```
fabianvf, line 92: @jortel 

I think this logic is implemented in the hub already (it's just using the normal auth flow, nothing solution server specific)

vscode/src/commands.ts
```diff
@@ -84,6 +84,23 @@ const commandsMap: (state: ExtensionState) => {
         console.error("Could not start the server", e);
       }
     },
+    "konveyor.getToken": async (): Promise<string | undefined> => {
+      let token = await state.extensionContext.secrets.get("konveyor.token");
+
+      if (!token) {
```
fabianvf, line 90: Just a note, if they don't have auth enabled in the hub we won't need to use a token to connect

PR #511, opened 2025-05-16
Feature/analysis fix as agent

pranavgaikwad:
This PR: 
- introduces a new filesystem cache for agents that can be invalidated / reset via extension to notify any changes on the disk
- adds a new workflow AnalysisIssueFix which moves the analysis issue fix generation logic into an agent: 
  - this is needed so that other agents are aware of any changes we made here


agentic/src/nodes/base.ts
```diff
@@ -257,12 +275,13 @@ export abstract class BaseNode extends KaiWorkflowEventEmitter {
     if (!this.tools || this.tools.length < 1 || !enableTools) {
       return response;
     }
+    const filteredTools = this.getToolsMatchingSelectors(toolsSelectors);
```
pranavgaikwad, line 278: we want each sub-agent to have access to only a specific set of tools. more tools != better work

djzager (approved):
ACK. I can't find any reason to hold this up and let others try it out.

fabianvf (approved):
LGTM

PR #562, opened 2025-07-01
:seedling: Bump kai version to v0.1.1-pre.0

feiskyer:
Without this, the server won't be able to start and would fail with error:

```
server args:
Analyzer RPC server started successfully [pid: 659950]
flag provided but not defined: -pipePath
```

### DIFF
--- a/scripts/collect-assets.js
+++ b/scripts/collect-assets.js
@@ -14,7 +14,7 @@ const cli = parseCli(
   {
     org: "konveyor",
     repo: "kai",
-    releaseTag: "v0.1.0",
+    releaseTag: "v0.1.1-pre.0",
     branch: "main",
     workflow: "build-and-push-binaries.yml",
     rulesetOrg: "konveyor",


djzager:
Thank you for the PR.

Developing from `main` you can collect assets from main like this:

image

You could be grabbing from the development-builds from https://github.com/konveyor/editor-extensions/releases/tag/development-builds and that should have the correct binaries included.

Could you articulate please what exactly causes this issue for you? I suspect we did something wrong.

feiskyer:
Here are the commands I'm running (those are commands from readme):

```
npm run build
npm run collect-assets
npm run dist
npm run package
```

So, without any parameters, npm would collect the assets from scripts/collect-assets.js, but the versions there are no longer compatible. So I think even we could use main branch manually, it's better to bump the version for compatibility?

feiskyer:
Yes, that is helpful. Let me close this one

PR #566, opened 2025-07-02
:bug: remove AWS_DEFAULT_REGION env validation

abrugaro:
Remove some validations that were making the extension to ignore if certain values were defined as system env variables

## Summary by CodeRabbit

* **Refactor**
  * Streamlined validation process by removing environment variable checks from model configuration validation.

### DIFF
--- a/vscode/src/client/modelProvider.ts
+++ b/vscode/src/client/modelProvider.ts
@@ -113,7 +113,6 @@ class ChatBedrockCreator implements ModelCreator {
 
   validate(args: Record<string, any>, env: Record<string, string>): void {
     validateMissingConfigKeys(args, ["model"], "model arg(s)");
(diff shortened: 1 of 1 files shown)


vscode/src/client/modelProvider.ts
```diff
@@ -113,7 +111,6 @@ class ChatBedrockCreator implements ModelCreator {
 
   validate(args: Record<string, any>, env: Record<string, string>): void {
     validateMissingConfigKeys(args, ["model"], "model arg(s)");
-    validateMissingConfigKeys(env, ["AWS_DEFAULT_REGION"], "environment variable(s)");
```
pranavgaikwad, line 116: can we only remove validation for aws as I know for sure others are good

pranavgaikwad (changes requested):

PR #584, opened 2025-07-10
:bug: update success rate more often

### LINKED ISSUE #583: Accepted status tag missing after rerunning analysis
### Description:

An incident that has previously gone through a successful modified + accepted cycle no longer shows the "accepted" tag after re-running the analysis. The expected behavior is that the tag  "1 accepted" should persist if the same incident is detected again.

**Steps to Reproduce:**

-  Run an analysis and modify + accept a suggested fix.

- Rerun the analysis without restarting the solution server.

- The incident appears again, but the accepted tag is missing.

**Workaround:**

Manually stopping and restarting the solution server restores the accepted tag from the database.

Screenshots:

Before restarting the server (tag missing):

Image

After restarting the server (tag visible):

Image

Expected Behavior:

The "1 accepted" tag should appear automatically after rerunning analysis, without needing to restart the server.

djzager:
Fixes #583


webview-ui/src/components/AnalysisPage/AnalysisPage.tsx
```diff
@@ -90,7 +90,7 @@ const AnalysisPage: React.FC = () => {
     if (enhancedIncidents.length > 0 && serverRunning) {
       dispatch(getSuccessRate());
     }
-  }, [enhancedIncidents.length, serverRunning, dispatch]);
+  }, [enhancedIncidents, serverRunning, dispatch]);
```
hhpatel14, line 93: Using **enhancedIncidents** runs the effect too often because the array reference keeps changing, even if the content doesn't. Results in too many requests and lead to a memory leak warnings. May be triggering this only when array length changes is a better option.

pranavgaikwad (approved):

PR #595, opened 2025-07-16
:sparkles: Unified logging in extension

### LINKED ISSUE #542: Standardize logging interface
One thing that became clear when attempting to help a user debug their setup and configuration was that our logging needed some love. The core logs are in the analyzer client but they simply are not verbose enough to give a full picture of what is happening. Additionally, we need logging capabilities across the project so simply using vscode outputChannel's is not sufficient.

One idea:
We could build a logger interface in the shared module that all components could implement and extend

```
export interface ILogger {
  debug(message: string, ...args: any[]): void;
  info(message: string, ...args: any[]): void;
  warn(message: string, ...args: any[]): void;
  error(message: string, ...args: any[]): void;
  child(options: LoggerOptions): ILogger;
}
```

1. build logger interface/factory/ core logging api
   1. Logger factory to manage and retrieve child loggers
1. Implement package specific loggers.
   1. Vscode
   1. webview
   1. agentic
1. Add log customization and log level etc. Begin migrating logs and testing

djzager:
With this change we can, throughout the extension codebase, use the logger to write to both the vscode output console AND a file in konveyor-logs in such a way that everything from commands, interactive workflows, interactions with the solution server, and even analyzer queries are logged as they happen.

Resolves #542 

## Summary by CodeRabbit

* **New Features**
  * Introduced a centralized, structured logging system across the extension using Winston, with logs written to both files and the VS Code output channel.
  * Added a new output channel name constant for improved log visibility within the editor.

* **Refactor**
  * Replaced all direct console logging with Winston-based structured logging in both extension and agent components.
  * Updated constructors and interfaces to accept and use a logger instance for consistent logging.

* **Chores**
  * Updated configuration options for logging levels and removed deprecated logging settings.
  * Added new dependencies to support structured logging.

### DIFF
--- a/vscode/src/utilities/logger.ts
+++ b/vscode/src/utilities/logger.ts
@@ -1,0 +1,50 @@
+import * as winston from "winston";
+import * as vscode from "vscode";
+import * as path from "path";
+import { ExtensionPaths } from "../paths";
+import { OutputChannelTransport } from "winston-transport-vscode";
+import { getConfigLogLevel } from "./configuration";
+import { KONVEYOR_OUTPUT_CHANNEL_NAME } from "@editor-extensions/shared";
+
+// Create logger instance
+export function createLogger(paths: ExtensionPaths): winston.Logger {
+  // Create or get the output channel
+  const outputChannel = vscode.window.createOutputChannel(KONVEYOR_OUTPUT_CHANNEL_NAME, "log");
+
+  // Create log file path
+  const logFile = path.join(paths.serverLogs.fsPath, "konveyor-extension.log");
+
+  const logger = winston.createLogger({
+    level: getConfigLogLevel(),
+    format: winston.format.combine(
+      winston.format.timestamp(),
+      winston.format.errors({ stack: true }),
+      winston.format.json(),
+    ),
+    transports: [
+      // File transport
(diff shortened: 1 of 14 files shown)


vscode/src/utilities/logger.ts
```diff
@@ -0,0 +1,49 @@
...
+    ),
+    transports: [
+      // File transport
+      new winston.transports.File({
+        filename: logFile,
+        maxsize: 10 * 1024 * 1024, // 10MB
+        maxFiles: 3,
+      }),
+      // VS Code output channel transport
+      new OutputChannelTransport({
```
pranavgaikwad, line 32: Can we set a different level for vscode transport? Thinking we keep the output limited to WARN / ERROR

pranavgaikwad (commented):
Overall looks good, just one question

vscode/src/utilities/logger.ts
djzager: Set the level in the vscode output channel to warn AND the console to "silly" for development.

pranavgaikwad (approved):
awesome work

ibolton336:
Looking good locally for me.

PR #608, opened 2025-07-22
:bug: Remove duplicate selection state for interaction messages

ibolton336:
- The yes/no buttons had some duplicate selection state logic. Leave the selection state to the underlying PF component. Use selectedResponse state for disabled, but let click handler indicate the selection.
Screenshot 2025-07-21 at 9 57 31 PM


djzager (approved):
ACK

PR #616, opened 2025-07-23
:bug: Fixes race conditions with the queue

ibolton336:
Fixes race conditions with the queue. Presented itself when:

Get solution request
blocking message received - pending interaction
Messages queued from stream that is continuous. Typical blocking queue wont really work here since we arent able to control flow of messages from llm in case of file messages. They just keep streaming.
Pending interaction resolved. 
Messages still streaming
Messages need to be processed for queue
^^race condition where messages get queued into an instance of queue list that is more up to date than the current queue execution context -> The messages get pushed to the queue and nothing ever processes them since processing has already been kicked off with a different queue list that does not contain the messages.

New PR changes - 

Give us the ability to just queue ALL messages up front - don’t toggle between processing and queueing based on is processing user interaction. ALWAYS queue first - prevents situations when queue messages can become jumbled and out of order. 

Additionally, if there is ever a race condition that is met in above scenario, added a message processor case to poll for queued messages that have not been processed as a last resort. With this new queue first / message buffering flow, I haven’t seen this happening but it is good to have as a failsafe. 

## Summary by CodeRabbit

* **Refactor**
  * Simplified and unified message queue handling for file modification workflows, enforcing consistent queue management and removing legacy fallback logic.
  * Increased logging detail for message processing and queue operations to aid in tracing and debugging.
  * Removed periodic monitoring of stuck interactions and redundant logging for pending interactions.
  * Streamlined control flow by requiring queue management for all relevant user interactions.
  * Introduced continuous background processing for queued messages with dynamic message handling and improved error logging.
  * Added cleanup and disposal mechanisms for queue management to ensure resource efficiency.
  * Replaced console logs with structured logging in file response handling for clearer error reporting and control flow.

### DIFF
--- a/vscode/src/utilities/ModifiedFiles/handleFileResponse.ts
+++ b/vscode/src/utilities/ModifiedFiles/handleFileResponse.ts
@@ -20,7 +20,9 @@ const createNewFile = async (
       try {
         await vscode.workspace.fs.createDirectory(directoryUri);
       } catch (dirError) {
-        console.error(`Failed to create directory at ${directoryPath}:`, dirError);
+        state.logger
+          .child({ component: "handleFileResponse.createNewFile" })
+          .error(`Failed to create directory at ${directoryPath}:`, dirError);
       }
     }
     await vscode.workspace.fs.writeFile(uri, Buffer.from(content));
@@ -28,7 +30,9 @@ const createNewFile = async (
       `Created new file ${vscode.workspace.asRelativePath(uri)}`,
     );
   } catch (error) {
-    console.error(`Failed to create file at ${filePath}:`, error);
+    state.logger
+      .child({ component: "handleFileResponse.createNewFile" })
+      .error(`Failed to create file at ${filePath}:`, error);
     throw new Error(`Failed to create file: ${error}`);
   }
 };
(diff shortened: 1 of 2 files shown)


vscode/src/commands.ts
```diff
@@ -205,7 +205,6 @@ const commandsMap: (
         state.modifiedFiles.clear();
         const modifiedFilesPromises: Array<Promise<void>> = [];
         // Queue to store messages that arrive while waiting for user interaction
-        const messageQueue: KaiWorkflowMessage[] = [];
```
ibolton336, line 208: Use the queue inside queueManager.ts

vscode/src/utilities/ModifiedFiles/handleFileResponse.ts
```diff
@@ -92,114 +87,101 @@ export async function handleFileResponse(
...
+      if (!resolved) {
+        console.warn(`No pending interaction found for messageToken: ${messageToken}`);
+        // As a fallback, reset the waiting flag if no pending interaction was found
+        // This should rarely happen if the architecture is working correctly
         state.isWaitingForUserInteraction = false;
       }
-    } finally {
-      console.log("Clearing loading state after file response handling");
+    } else {
+      console.warn(
```
pranavgaikwad, line 176: change console.warn to logger?

vscode/src/utilities/ModifiedFiles/queueManager.ts
```diff
@@ -50,9 +45,39 @@ export class MessageQueueManager {
...
+
+    this.processingTimer = setInterval(() => {
+      // Only process if we're not already processing and not waiting for user
+      if (
+        !this.isProcessingQueue &&
+        !this.state.isWaitingForUserInteraction &&
+        this.messageQueue.length > 0
+      ) {
+        this.processQueuedMessages().catch((error) => {
+          console.error("Error in background queue processing:", error);
```
pranavgaikwad, line 62: same...use logger?

pranavgaikwad (changes requested):
I tested this, and I can confirm this solves the race conditions in the queue. Nits around changing console.log to logger instead.

PR #626, opened 2025-07-26
:sparkles: Implement Profile Deletion and Test Cleanup

RanWurmbrand:
**Description**
This PR adds the functionality to delete analysis profiles and integrates it into the e2e test suite to ensure tests clean up after themselves.

**Implementation Details**

**1. New Deletion Method:**
- A deleteProfile method was added to the page object, automating the UI flow for deleting a profile.

**2. E2E Test Integration:**
- The configure-and-run-analysis.test.ts file was updated to verify this new feature.
- The test now generates a unique profile name during setup.
- A new test case was added that calls deleteProfile with this unique name, confirming the end-to-end deletion flow and ensuring no artifacts are left behind.

### DIFF
--- a/tests/e2e/pages/vscode.pages.ts
+++ b/tests/e2e/pages/vscode.pages.ts
@@ -213,7 +213,7 @@ export class VSCode extends Application {
     await manageProfileView.getByRole('button', { name: '+ New Profile' }).click();
 
     const randomName = generateRandomString();
-    const nameToUse = profileName ? `${profileName}-${randomName}` : randomName;
+    const nameToUse = profileName ? profileName : randomName;
     await manageProfileView.getByRole('textbox', { name: 'Profile Name' }).fill(nameToUse);
 
     // Select Targets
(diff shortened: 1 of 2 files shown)


tests/e2e/pages/vscode.pages.ts
```diff
@@ -213,7 +213,7 @@ export class VSCode extends Application {
     await manageProfileView.getByRole('button', { name: '+ New Profile' }).click();
 
     const randomName = generateRandomString();
-    const nameToUse = profileName ? `${profileName}-${randomName}` : randomName;
+    const nameToUse = profileName ? profileName : randomName;
```
sshveta, line 216: Why are we changing this ? 
The intention is to add randomstring after ProfileName always so that there is no duplicate even if the profile failed to delete .

RanWurmbrand: We've moved the random string generation to the caller (the test function).

This allows us to create the full profile name within the describe block, making it accessible to other tests that need it (e.g., the delete profile test).

The alternative, deleting the profile via regex, was problematic. It required adding flawed default behaviors to handle edge cases, such as:

- When no profile name was provided.

- When a name prefix matched multiple profiles.

This approach avoids those complexities and prevents the accidental deletion of unintended profiles.

sshveta (approved):

PR #633, opened 2025-07-29
:bug: Fix type errors in dev spaces

ibolton336:
In DevSpaces, there's often a race condition where React components mount before the extension fully initializes the webview data. 

There is also a CDN pf font that is unreachable in prod mode. This is now overridden by local PF fonts.


djzager (approved):
VISACK

PR #727, opened 2025-08-28
:bug: analyzer and genai config fixes

### LINKED ISSUE #720: The server can be started with a non existing analyzer binary
If the user overrides the analyzer path in the configuration with a path that doesn’t exist, then tries to start the server, an error notification is displayed and the loading spinner does not appear. 

However, the server still starts after a few seconds and every analysis will fail.

https://github.com/user-attachments/assets/0202274f-d48f-4db4-9b50-7ef745f2cf49

The server should not start if the analysis binary path doesn't exist

Affects #712 

Env:

* Latest development build
* Fedora 41

djzager:
This addresses a few issues:

1. #720 - updated the ensureKaiAnalyzerBinary to handle the user configured and throw vscode error if invalid (and reset)...use it for startup and config changes
2. Only prompt for ss stuff if auth is enabled AND solution server is enabled
3. Fix UX for disabled gen ai case.
Fixes #720 

## Summary by CodeRabbit

- New Features
  - Enable GenAI directly from alerts, the walkthrough step, or via a new command.
- Improvements
  - Clear “GenAI is disabled” status with guided actions to enable.
  - Analyzer path setting now validates user-provided paths, falls back to bundled binary if invalid, and restarts the analyzer when the path changes.
  - Solution server connects only when enabled, reducing noise.
- Bug Fixes
  - Prevents unnecessary credential prompts when the solution server is disabled.
- Chores
  - Minor UI cleanup and formatting; removed an unused import.

### DIFF
--- a/shared/src/types/actions.ts
+++ b/shared/src/types/actions.ts
@@ -3,6 +3,7 @@ export const RUN_ANALYSIS = "RUN_ANALYSIS";
 export const START_SERVER = "START_SERVER";
 export const STOP_SERVER = "STOP_SERVER";
 export const RESTART_SOLUTION_SERVER = "RESTART_SOLUTION_SERVER";
+export const ENABLE_GENAI = "ENABLE_GENAI";
 export const CANCEL_SOLUTION = "CANCEL_SOLUTION";
 export const GET_SOLUTION = "GET_SOLUTION";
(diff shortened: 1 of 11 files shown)


fabianvf (approved):
VISACK

Issue #742, opened 2025-08-30
(BUG) Incorrect Profile Deletion Behavior


RanWurmbrand:
**Description**
When a user attempts to delete a profile using the kebab menu, the application does not delete the intended profile. Instead, the deletion action consistently affects the last profile in the list, regardless of which profile's "Delete" button was clicked.

**Steps to Reproduce**

1. Open the Profile List component with at least two profiles.
2. Click the kebab menu (`...`) for any profile other than the last one in the list.
3. From the dropdown menu, select "Delete".
4. Observe that a confirmation dialog appears, but the message incorrectly references the last profile in the list.
5. Click the "Delete" button in the confirmation dialog.
6. The last profile is removed from the list, while the intended profile remains.

**Root Cause**
While attempting to confirm the reason for the incorrect deletion, a key behavioral observation was made: when the "Delete" button is clicked for a specific profile, an instance of the confirmation dialog is unexpectedly created and opened in the DOM for every profile in the list, not just the one that was selected.

https://github.com/user-attachments/assets/f445a4fc-685e-49ab-b351-553a36b206dc

**Proposed Solution**
The `ConfirmDialog` component needs to be moved from inside the `profiles.map()` loop to a position outside of it. A single instance of `ConfirmDialog` should be rendered and its state (`isOpen` and `profileToDelete`) must be managed centrally. This will ensure that only one dialog exists at a time, and its content is correctly tied to the profile selected for deletion by the `onClick` handler of the "Delete" button. This change will properly link the deletion action to the specific profile selected by the user.

abrugaro:
Duplocate of https://github.com/konveyor/editor-extensions/issues/565

PR #743, opened 2025-08-30
:bug: Refactor profile deletion dialog logic

RanWurmbrand:
feat Incorrect Profile Deletion Behavior #742

The changes address an unintended side effect where multiple dialog instances were being rendered, leading to incorrect profile deletion. The fix consists of two parts:

- **Centralized Dialog Rendering**: The `ConfirmDialog` component was moved outside the `profiles.map()` loop. This change ensures only one dialog instance is ever present in the DOM, preventing the rendering of multiple hidden dialogs.

- **Explicit State Management**: The state variable for managing the dialog's visibility was updated from a simple boolean to a `profileToDelete` object. This change provides a direct and explicit link between the dialog and the specific profile selected for deletion, ensuring the correct profile is targeted every time.


ibolton336 (approved):
LGTM!

Issue #745, opened 2025-09-01
Can't connect to solution server deployed with Konveyor operator


abrugaro:
While trying to access an authenticated solution server deployed with the konveyor operator I get the following error.

I know the authentication is working and the URL is correct because if I disable it I get a 401 error.
```
{"component":"SolutionServerClient","level":"info","message":"Disconnecting from MCP solution server...","timestamp":"2025-08-29T16:15:22.207Z"}
{"component":"SolutionServerClient","level":"debug","message":"Restoring SSL settings","timestamp":"2025-08-29T16:15:22.207Z"}
{"component":"SolutionServerClient","level":"info","message":"Disconnected from MCP solution server","timestamp":"2025-08-29T16:15:22.207Z"}
{"cause":{"code":"ENOTFOUND","errno":-3008,"hostname":"TheProject-hub.openshift-TheProject.svc","syscall":"getaddrinfo"},"level":"error","message":"Error connecting to solution server fetch failed","stack":"TypeError: fetch failed\n\tat node:internal/deps/undici/undici:13510:13\n\tat process.processTicksAndRejections (node:internal/process/task_queues:105:5)\n\tat async globalThis.fetch (file:///usr/share/code/resources/app/out/vs/workbench/api/node/extensionHostProcess.js:166:24322)\n\tat async Su.send (/home/abrugaro/.vscode/extensions/konveyor.konveyor-0.2.0-pre.2/out/extension.js:2:1823952)","timestamp":"2025-08-29T16:15:22.208Z"}
{"component":"webviewMessageHandler","level":"debug","message":"messageHandler: {\"type\":\"WEBVIEW_READY\"}","timestamp":"2025-08-29T16:15:22.838Z"}
{"component":"webviewMessageHandler","level":"info","message":"Webview is ready","timestamp":"2025-08-29T16:15:22.838Z"}
```

ibolton336:
I was also able to reproduce this with similar results 

[...]

PR #751, opened 2025-09-02
:bug: Retry logic for connection attempts with and without trailing slash

ibolton336:


fabianvf (approved):
We should file an issue to fix the underlying issue+remove this

ibolton336:
> We should file an issue to fix the underlying issue+remove this

Added a comment [here](https://github.com/konveyor/editor-extensions/issues/745#issuecomment-3247153413) to track against the original issue.

ibolton336:
Closes https://github.com/konveyor/editor-extensions/issues/745

PR #756, opened 2025-09-03
:bug: Manage profiles form validation fix

ibolton336:
Screenshot 2025-09-03 at 12 48 28 PM

Adds form validation for the manage profiles page. 

TODO: There are still cases where the configuration.ts server side error notification creation engine is failing to push notifications for:

```
export function updateConfigErrors(draft: ExtensionData, settingsPath: string): void {
  const { activeProfileId, profiles } = draft;
  const profile = profiles.find((p) => p.id === activeProfileId);

  // Check for no active profile
  if (!profile) {
    draft.configErrors.push(createConfigError.noActiveProfile());
    return;
  }

  // Check label selector
  if (!profile.labelSelector?.trim()) {
    draft.configErrors.push(createConfigError.invalidLabelSelector());
  }

  // Check custom rules when default rules are disabled
  if (!profile.useDefaultRules && (!profile.customRules || profile.customRules.length === 0)) {
    draft.configErrors.push(createConfigError.noCustomRules());
  }
}

```

This updateConfigErrors approach should be improved to better sync with UI state. 

## Summary by CodeRabbit

- New Features
  - Added cross-field validation in Profile Editor for Targets and Rules.
  - Real-time checks run on load and whenever related fields change.
  - Contextual error messages replace static helper text; guidance shown when valid.
  - Inline icons highlight errors vs. info for clearer feedback.
  - Form-level alert appears when Name, Targets, or Rules are invalid.
  - Ensures at least one target and at least one rules configuration (default or custom) before proceeding.
  - Revalidates after label selector updates and when custom rules are removed.

### DIFF
--- a/webview-ui/src/components/ProfileManager/ProfileEditorForm.tsx
+++ b/webview-ui/src/components/ProfileManager/ProfileEditorForm.tsx
@@ -68,6 +68,12 @@ export const ProfileEditorForm: React.FC<{
   const [nameValidation, setNameValidation] = useState<"default" | "error">("default");
   const [nameErrorMsg, setNameErrorMsg] = useState<string | null>(null);
 
+  const [targetsValidation, setTargetsValidation] = useState<"default" | "error">("default");
+  const [targetsErrorMsg, setTargetsErrorMsg] = useState<string | null>(null);
+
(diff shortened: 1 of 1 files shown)


djzager (approved):
visack

PR #757, opened 2025-09-03
:bug: Clear auth config when authentication is disabled

### LINKED ISSUE #773: Auth config is still passed when solution server auth.enabled is not explicitly set to true.
Fix issue where realm parameter was still passed to solution server connection even when auth is not set. Should clear the auth configuration when authentication is not enabled.

ibolton336:
Fix issue where realm parameter was still passed to solution server connection even when auth is turned off. Now explicitly clears the auth configuration when authentication is disabled.

Resolves #773 

## Summary by CodeRabbit

* **Bug Fixes**
  * Clears solution-server authentication settings when authentication is disabled or the server is turned off.
  * Stops sending authentication data or headers when no auth configuration or token is available.
  * Ensures token exchange and related auth steps only run when auth configuration is present.
  * Preserves existing behavior when valid credentials are supplied: configuration is set on success and errors are reported on failure.

### DIFF
--- a/agentic/src/clients/solutionServerClient.ts
+++ b/agentic/src/clients/solutionServerClient.ts
@@ -78,6 +78,13 @@ export class SolutionServerClient {
 
   public setAuthConfig(authConfig: SolutionServerAuthConfig | null): void {
     this.authConfig = authConfig;
+
+    // Clear auth tokens and timers when auth is disabled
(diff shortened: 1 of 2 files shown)


ibolton336:
TIL: We explicitly need auth enabled: false to prevent auth config being passed if previously enabled.

PR #847, opened 2025-09-17
:bug: only add ss-auth configerror if not there already

### LINKED ISSUE #841: Authentication warning persists and duplicates when toggling Solution Server
### Description:

I launched vscode and added configured the solution server:

```
        "auth": {
            "enabled": true,
            "realm": "mta",
            "insecure": true,
            ...
        },
```

on the Analysis view page it showed me a warning: `Authentication is enabled but credentials are not configured.` despite the username and password being configured. so I then disabled the authentication but the warning persisted, and then re-enabled the auth yet again, but this time the warning duplicated

Image

I only observed it once, after restarting the vscode I couldn't find a clear reproducing steps.

Environment:
OS: Mac OS
build: [mta-vscode-extension-0.8.0-beta.1.vsix](https://github.com/migtools/editor-extensions/releases/download/v0.8.0-beta.1/mta-vscode-extension-0.8.0-beta.1.vsix)

djzager:
Fixes #841

### DIFF
--- a/vscode/src/extension.ts
+++ b/vscode/src/extension.ts
@@ -609,7 +609,9 @@ class VsCodeExtension {
       const credentials = await checkAndPromptForCredentials(this.context, this.state.logger);
       if (!credentials) {
         this.state.mutateData((draft) => {
-          draft.configErrors.push(createConfigError.missingAuthCredentials());
+          if (!draft.configErrors.some((error) => error.type === "missing-auth-credentials")) {
+            draft.configErrors.push(createConfigError.missingAuthCredentials());
+          }
         });
         return;
       }


djzager:
Adding do not merge while we are in this code freeze.

vscode/src/extension.ts
```diff
@@ -609,7 +609,9 @@ class VsCodeExtension {
       const credentials = await checkAndPromptForCredentials(this.context, this.state.logger);
       if (!credentials) {
         this.state.mutateData((draft) => {
-          draft.configErrors.push(createConfigError.missingAuthCredentials());
+          if (!draft.configErrors.some((error) => error.type === "missing-auth-credentials")) {
```
pranavgaikwad, line 612: I think this is good for now but filing an issue to make configErrors a managed set so we don't always to have make these checks

pranavgaikwad (approved):

PR #855, opened 2025-09-19
:bug: Handle max_tokens in Bedrock responses

### LINKED ISSUE #763: Handle Models Returning early instead of full code block
When a model returns without the full code, but instead has `(Rest Of Code Stays the Same)` or something similar, then we should consider handling this as it says. Today, we take that as an actual change, most likely leading to many solutions that may have been accepted but need to be modified.

### LINKED ISSUE #856: When applying a solution to an incident, Java files are sometimes modified incorrectly, leaving the text "The file remains unchanged" in the code.
**Steps to Reproduce**

Open an incident in the extension.
Apply the suggested solution to the incident.
Check the modified Java files in the project.

Image

pranavgaikwad:
Fixes #856
Fixes #763 

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated Bedrock model provider with token-limit-aware handling for streaming and non-streaming chats.
  * Analysis fix flow now returns the full updated file output for complete results.

* **Improvements**
  * Unified provider construction via a configurable options object for more consistent provider setup.
  * More reliable message continuations near token limits with consistent caching and tracing.

* **Bug Fixes**
  * Fixed partial-response/continuation issues under max-token constraints.

* **Tests**
  * Updated cached test data and metadata for broader coverage and stability.

### DIFF
--- a/agentic/src/types.ts
+++ b/agentic/src/types.ts
@@ -1,8 +1,4 @@
-import {
-  type AIMessageChunk,
-  type AIMessage,
-  type BaseMessageChunk,
-} from "@langchain/core/messages";
+import { type AIMessageChunk, type AIMessage, type BaseMessage } from "@langchain/core/messages";
 import {
   type BaseChatModelCallOptions,
   type BindToolsInput,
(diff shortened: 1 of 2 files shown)


djzager (approved):
LGTM

PR #946, opened 2025-10-22
:test_tube: Solution Server Settings Test case

sshveta:
## Summary by CodeRabbit

* **Tests**
  * Added end-to-end tests for Solution Server covering auth, realm, and security permutations.
  * Updated generative-AI test flows to write workspace settings via an in-editor JSON settings helper and added a reusable settings-write utility.
  * Added a configuration key to toggle GenAI.

* **Documentation**
  * Documented four new environment variables for Solution Server configuration.

* **Style**
  * Clarified wording in the environment example comments.


tests/e2e/pages/vscode.page.ts
```diff
@@ -1,43 +1,198 @@
...
     await this.waitDefault();
     const modifier = getOSInfo() === 'macOS' ? 'Meta' : 'Control';
     await this.window.keyboard.press(`${modifier}+Shift+P`, { delay: 500 });
     const input = this.window.getByPlaceholder('Type the name of a command to run.');
     await input.fill(`>${command}`);
-    await expect(
-      this.window.locator(`a.label-name span.highlight >> text="${command}"`)
-    ).toBeVisible();
-
+    await expect(this.window.locator('a').filter({ hasText: command })).toBeVisible({
```
abrugaro: `hasText` searches for substring as well so this may not work when there are 2 similar commands, like:

```
Konveyor: Open the default the GenAI model provider configuration file and backup the current file
Konveyor: Open the GenAI model provider configuration file
```

tests/e2e/tests/solution-server/settings.test.ts
```diff
@@ -0,0 +1,119 @@
...
+  password?: string;
+};
+
+const solutionServerConfigs: SolutionServerConfig[] = [
+  {
+    name: 'All enabled',
+    ssEnabled: true,
+    authInIDE: true,
+    insecure: true,
+    realm: 'mta',
```
abrugaro: `realm`, `user` and `password` should be provided by env variables. I think @RanWurmbrand  was working on enabling solution server configuration with env variables so he may update this later.


Also `realm` should default to `konveyor`.

abrugaro (changes requested):
Some minor things

tests/e2e/tests/solution-server/settings.test.ts
abrugaro: Yes, right!

tests/e2e/tests/solution-server/settings.test.ts
```diff
@@ -0,0 +1,119 @@
...
+      await vscodeApp.waitDefault();
+
+      if (scenario.shouldConnect) {
+        await expect(
+          analysisView.getByRole('heading', { name: 'Warning alert: Solution' })
+        ).not.toBeVisible();
+      } else {
+        await expect(
+          analysisView.getByRole('heading', { name: 'Warning alert: Solution' })
+        ).toBeVisible();
```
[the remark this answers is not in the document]
abrugaro: In the documentation it states that the boolean argument `visible` can be passed as an option.

https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-visible

But I'm totally ok with leaving it as it is, maybe it doesn't work because the element is not attached

abrugaro (approved):

Issue #947, opened 2025-10-22
Bug: Fix all issues in non-agent mode hangs


ibolton336:
https://github.com/user-attachments/assets/31b27c13-7d0c-405a-ad3e-5fe8ed7420c0

Issue #888, opened 2025-10-01
[Windows] Analysis View tab is duplicated afer accepting some solution proposals


abrugaro:
After aceepting 4 or 5 solution proposals, the resolution view opens automatically a new Analyis view while keeping the other one open as well. The original Analysis view has the server running and its stuck on "waiting for user action" while the new one has the server stopped.

**Steps to reproduce**

1. Analyze the coolstore app
2. Click the fix all issues button
3.  Accept a couple of proposals.

<img width="720" height="482" alt="Image" src="https://github.com/user-attachments/assets/7644aa4b-be8f-433b-bac8-dd3c921bcf81" />

Env:

* Windows 2022 Datacenter running on an ec2 c4.xlarge in AWS
* Extension built from main branch
* Coolstore app

abrugaro:
This error is no longer reproducible

PR #957, opened 2025-10-24
:bug: Handle early cleanup bug with fix-all flows

ibolton336:
[Bug: Fix all issues in non-agent mode hangs](https://github.com/konveyor/editor-extensions/issues/947)
https://github.com/konveyor/editor-extensions/issues/888

Currently, "fix all" has a race condition where the workflow will complete while messages are still in queue. The finally block of getSolution cleans up the queue manager and leaves the webview in a permanent loading state. 

This PR aims to improve the getSolution method in commands by moving away from the nested try/catch blocks to a more functional approach. 

## Summary by CodeRabbit

* **New Features**
  * Shows a "Processing solution..." status and backdrop while queued messages are processed
  * Adds an "Open Resolution Panel" button on the backdrop to open the resolution UI

* **Improvements**
  * Clearer loading indicators reflecting combined processing and user-interaction states
  * Pending user interactions auto-resolve after a timeout to prevent stalls
  * More reliable background queue processing, error recovery, and UI state visibility during workflows

### DIFF
--- a/vscode/core/src/solutionWorkflowOrchestrator.ts
+++ b/vscode/core/src/solutionWorkflowOrchestrator.ts
@@ -1,0 +1,556 @@
+import * as vscode from "vscode";
+import { ExtensionState } from "./extensionState";
+import { EnhancedIncident, Scope, ChatMessageType } from "@editor-extensions/shared";
+import {
+  type KaiWorkflowMessage,
+  type KaiInteractiveWorkflowInput,
+} from "@editor-extensions/agentic";
+import { getConfigAgentMode } from "./utilities/configuration";
+import { executeExtensionCommand, executeDeferredWorkflowDisposal } from "./commands";
+import { processMessage } from "./utilities/ModifiedFiles/processMessage";
+import { MessageQueueManager } from "./utilities/ModifiedFiles/queueManager";
+import { v4 as uuidv4 } from "uuid";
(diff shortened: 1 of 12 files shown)


vscode/core/src/solutionWorkflowOrchestrator.ts
```diff
@@ -0,0 +1,556 @@
...
+
+    // Wait for processing based on mode
+    await this.waitForProcessing();
+  }
+
+  /**
+   * Wait for message processing to complete based on agent mode
+   */
+  private async waitForProcessing(): Promise<void> {
+    if (this.agentMode) {
```
ibolton336: Candidate for removal. Just carrying over from previous implementation.

vscode/core/src/solutionWorkflowOrchestrator.ts
```diff
@@ -0,0 +1,556 @@
...
+        workflowRunCompleted: this.workflowRunCompleted,
+        isFetchingSolution: this.state.data.isFetchingSolution,
+      });
+    }
+  }
+
+  /**
+   * Create the resolver function for pending interactions
+   */
+  private createInteractionResolver() {
```
ibolton336: Moved the resolver function here for when accept/reject is handled in one way or another in modified file msg flow.

Issue #969, opened 2025-10-30
🧪 verify issues reported for the analysis


midays:
verifiy the subset of issues reported for each analysis

PR #994, opened 2025-11-09
:test_tube: verify issues reported for the analysis

midays:
Solves https://github.com/konveyor/editor-extensions/issues/969

## Summary by CodeRabbit

* **Tests**
  * Enhanced end-to-end tests to collect and verify issue cards (title + incident counts).
  * Test fixtures and types updated to include per-repo issues arrays; assertions now validate issue lists and counts.

* **Chores**
  * Removed a redundant wait during test initialization.
  * Minor workflow formatting consistency changes and a small dependency version bump.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,19635 @@
+{
+  "name": "editor-extensions",
+  "version": "0.4.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "editor-extensions",
+      "version": "0.4.0",
+      "hasInstallScript": true,
+      "license": "Apache-2.0",
+      "workspaces": [
+        "extra-types",
+        "shared",
+        "webview-ui",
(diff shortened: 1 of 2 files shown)


abrugaro (dismissed):
LGTM! please test it on windows before merging

tests/global.setup.ts
```diff
@@ -24,7 +24,7 @@ async function globalSetup() {
   await vscodeApp.closeVSCode();
   console.log('Completed global setup.');
 
-  if (getOSInfo() === 'windows' && process.env.CI) {
+  if (getOSInfo() === 'windows' || process.env.CI) {
```
abrugaro: This will also add extra step to the linux pipeline, which can run faster and don't need this.

When we're running the tests locally on Windows we can always update these values so therei s no need to commit them.

The same applies to the previous change

midays: done

PR #1036, opened 2025-11-20
:bug: (backport) Address ECONNRESET when http2 blocked

ibolton336:


jmontleon (approved):
ACK. Tested and works as expected for me.

Issue #913, opened 2025-10-08
Newest version of Java extension causes false positives during analysis


fabianvf:
Related to konveyor/analyzer-lsp#901

Partially addressed in #911 , which notifies the user when the incompatibility will occur. 

We need to
1. Determine how far reaching the impact is (is #of rules, # of false positives in our test project
2. Find the root cause in the LS and have it fixed
3. In the meantime, investigate if there's a postprocessing step we can add to the bundle, analyzer, or extension to filter out these false positives

PR #1037, opened 2025-11-21
:ghost: Remove JAVA version warning from release-0.2

ibolton336:


jwmatthews (approved):
VISIACK

This work is related to https://github.com/konveyor/editor-extensions/issues/913, as of today we have not been able to replicate the concern with Java Language Server 1.46.0+.  Thinking is to remove the warning to minimize confusion to users.

If we end up replicating the concern in issue #913 we can put the warning back in until we have a proper fix.

PR #1070, opened 2025-12-03
:sparkles: activate language extensions on workspace files

djzager:
The goal here is to make it so that each of our language specific extensions activates themselves when the workspace contains their language files. This should force them to register themselves with the konveyor-core extension ('konveyor.konveyor') and then when startup is complete we should be safe to start the analyzer server with some confidence.

## Summary by CodeRabbit

* **Refactor**
  * Java, Go, and JavaScript support now activates automatically based on detected project files and language events (e.g., build manifests, module files, package/config files) instead of loading at startup — improving startup performance and loading only when relevant.
  * Explicit manual trigger to open a Java file for activation has been removed; activation relies on workspace/language detection.

### DIFF
--- a/tests/e2e/pages/vscode.page.ts
+++ b/tests/e2e/pages/vscode.page.ts
@@ -727,30 +727,4 @@ export abstract class VSCode {
 
     return results;
   }
-
-  /**
-   * Opens a Java file to trigger onLanguage:java activation for both redhat.java and konveyor-java
-   */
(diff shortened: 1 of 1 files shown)


savitharaghunathan (approved):
visiack

pranavgaikwad (approved):
Thank you, this is much cleaner.

pranavgaikwad (approved):

PR #1075, opened 2025-12-05
:sparkles: Added health check command

JonahSussman:
## Summary by CodeRabbit

* **New Features**
  * Added a "Run Health Check" command to execute a comprehensive diagnostics workflow with progress, output logging, and optional report saving/opening.
  * Built-in health checks for analyzer binary/server, filesystem permissions, named-pipe/socket support, network/proxy/TLS, LLM provider connectivity, Windows security signals, Java tooling, and more.
  * Introduced an extensible registry/API so extensions can register custom health checks.

### DIFF
--- a/vscode/core/src/healthCheck/index.ts
+++ b/vscode/core/src/healthCheck/index.ts
@@ -97,9 +97,9 @@ export async function runHealthCheck(
 
   for (const module of applicableModules) {
     logger.debug(`Running health check: ${module.name}`);
+    // Track duration for each individual health check
+    const checkStartTime = performance.now();
     try {
-      // Automatically track duration for each health check
-      const checkStartTime = performance.now();
       const result = await module.check(context);
       const duration = performance.now() - checkStartTime;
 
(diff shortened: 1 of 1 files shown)


vscode/core/src/healthCheck/checks/javaExtension.ts
```diff
@@ -0,0 +1,74 @@
+/**
+ * Health check for Java Language Server (Red Hat Java extension)
+ */
+
+import { HealthCheckModule, CheckResult, HealthCheckContext } from "../types";
+
+export const javaExtensionCheck: HealthCheckModule = {
```
djzager: This one doesn't really make sense in our multi-language extension world. It's totally valid to not have the java extension installed.

Maybe what we could do is delegate this to the installed and registered extensions? Like maybe we extend the registration api to include health check callbacks? Might be beyond the scope of this PR but I don't think we should include this as it is right now.

djzager (commented):
Many of these health check modules, in my eyes, are ripe for simplification to reduce their total line count.

I am struggling to reconcile how to feel about this knowing this was mostly AI just cranking out tokens but there is an interesting venn diagram of tests where, if the failure condition was met then the extension wouldn't activate and the command to run this health report wouldn't be available, to ones that we thought were the source of odd bugs but turned out not to be.

I am certain there are legit tests in here and I'm okay with covering the whole spectrum but we should at least consider minifying.

pranavgaikwad (commented):
This is looking good, I have two minor concerns...

pranavgaikwad (approved):
This already does a lot, I am good with doing follow up fixes.

djzager (approved):

PR #1121, opened 2025-12-23
:bug: Enable skip ssl toggle even if auth is disabled

### LINKED ISSUE #1120: Disabling the authentication in the hub configuration also disables Skip SSL toggle
If a user disables the Authentication toggle in the hub configuration page, the Skip SSL toggle also becomes disabled. This leads to an error when Konveyor is deployed with authentication disabled and a self-signed certificate, since there is no way to skip SSL verification.

**Steps to Reproduce**
1. Go to the hub configuration page.
2. Disable the Authentication toggle.
3. Notice that the Skip SSL toggle becomes disabled as well.
4. Deploy Konveyor with authentication disabled and a self-signed certificate.
5. An error occurs due to SSL verification with no way to skip it.

**Expected behavior**
Disabling authentication should not disable the Skip SSL toggle. Users should be able to deploy with authentication disabled and still skip SSL verification for self-signed certificates.

abrugaro:
Resolves #1120 

## Summary by CodeRabbit

* **New Features**
  * Moved SSL certificate verification control into General Settings for clearer organization.
  * "Skip SSL certificate verification" can now be toggled independently of authentication state.
  * Helper text now shows a clear warning when the option is enabled and a generic caution when disabled.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3088,7 +3088,9 @@
       }
     },
     "node_modules/@langchain/core": {
-      "version": "0.3.66",
+      "version": "0.3.80",
+      "resolved": "https://registry.npmjs.org/@langchain/core/-/core-0.3.80.tgz",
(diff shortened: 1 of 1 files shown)


djzager (changes requested):
Thank you for this. Fix the lint warning and we can merge.

djzager (dismissed):

PR #1188, opened 2026-01-16
:bug: analysis should ignore what git ignores

### LINKED ISSUE #1182: Analyzing tackle2-ui with patternfly rules appears to hang if project was previously built
I tested the konveyor-javascript extension v0.3.4 and v0.3.11 with:
- Sample App: https://github.com/konveyor/tackle2-ui/tree/main
- Rules: https://github.com/konveyor/rulesets/tree/main/preview/nodejs/patternfly
- Environment MacOS arm64

I saw the progress bar move quick until it got up to ~224/230 rules processed, then I saw the analysis continue for 24+ hours and still not complete.

Was able to narrow the issue down to a handful of `perl` processes were running such as `perl -ne if (/import[\s\S]*DualListSelector[\s\S]*from[\s\S]*['"]@patternfly\/react-core\/next['"]/) { print "$ARGV:$.:$1\n" } `

Looking at the command line of one of these perl processes as an example you can see ~400 files are passed in on the commandline for perl to process:  https://gist.githubusercontent.com/jwmatthews/ee4c077465a1d91a7566f6c2c95d6d29/raw/3484bd313b32ad996b9c399070de421b5745ae98/gistfile1.txt

I was able to identify that this is the first file that presented a problem to process a 11MB file `tackle2-ui/client/dist/app.bundle.js`

## Thoughts
I think the first consideration is it looks like while there is a .gitignore in the parent directory of tackle2-ui that has an entry for `dist/` I suspect this is not applying to the sub directories, so the other sub-projects like `client` which have a `client/dist` are not being ignored and are being processed.

Second thought, this use-case may serve as a decent stress test if we wanted to explore improving the builtin searches so they could handle even more difficult situations. 

## Reproducer and workaround

1. `git clone https://github.com/konveyor/tackle2-ui.git`
2. `cd tackle2-ui`
3. `npm install`
4. `npm run build`
5. Now run an analysis from the IDE with konveyor-javascript and the new patternfly rules
6. See how ~223 rules process in <2 mins, then CPU usage spikes as several perl processes run, in my case multiple hours passed and I still had ~4-5 perl processes running, only ~2 of them completed.
7. Kill the analysis
8. Clean things up, `npm run clean`
9. Verify you do NOT see a dist directory under `client/dist`
10. Re-run analysis, see it complete in a few minutes

djzager:
Fixes #1182

## Summary by CodeRabbit

* **Refactor**
  * Improved ignore-file handling for more consistent, reliable exclusion of files and directories (priority-based ignore loading, normalized path checks, and stable defaults).

* **New Features**
  * Workspace ignore resolution now prefers repository-specific ignore files, falls back to sensible defaults, and always omits internal config folders.

* **Tests**
  * Added comprehensive tests covering pattern semantics, negations, rooted patterns, and realistic ignore-file scenarios.

### DIFF
--- a/vscode/core/src/paths.ts
+++ b/vscode/core/src/paths.ts
@@ -1,8 +1,7 @@
 import { relative, dirname, join } from "node:path";
-import { readFileSync, createWriteStream, createReadStream } from "node:fs";
+import { createWriteStream, createReadStream } from "node:fs";
 import { globbySync, isIgnoredByIgnoreFilesSync } from "globby";
 import * as vscode from "vscode";
-import slash from "slash";
 import winston from "winston";
 import { createHash } from "node:crypto";
 import { mkdir, chmod, unlink } from "node:fs/promises";
@@ -11,7 +10,11 @@ import { platform, arch } from "node:process";
 import { existsSync } from "node:fs";
 import { getConfigAnalyzerPath } from "./utilities/configuration";
 import { EXTENSION_NAME } from "./utilities/constants";
-import { parseIgnoreFileToGlobPatterns } from "./utilities/ignorePatterns";
+import {
+  createIgnoreFromWorkspace,
+  filterIgnoredPaths,
+  DEFAULT_IGNORE_PATTERNS,
+} from "./utilities/ignorePatterns";
 import AdmZip from "adm-zip";
 
 /**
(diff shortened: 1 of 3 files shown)


djzager:
At this stage we should be able to see the breaking code pulled out into a utility so we can properly test it without pulling in globby/unicorn/vscode that causes lots and lots of headaches.

Issue #565, opened 2025-07-02
Deleting a profile always deletes the last one


abrugaro:
When the user deletes a profile, the extension always deletes the last one in the list no matter the one selected by the user.

https://github.com/user-attachments/assets/7863362d-8902-4b27-9593-01841cac3e96

Tested on development build https://github.com/konveyor/editor-extensions/releases/download/development-builds/konveyor-ai-20250702-0554-b8a0380.vsix

PR #1285, opened 2026-03-01
:test_tube: Fix profile e2e test issues

RanWurmbrand:
### Profile name fix
The profile name `profileWithRules-<random>` exceeded the 24-character limit enforced by the sidebar. When a profile name is too long, the sidebar truncates it to 24 characters, which caused the locator to fail when searching for the full name or `<name> (active)`. Renamed the prefix to `rulesProfile` to keep it within the limit.

### Unskip delete profile test
The "Delete profile using action Button" test was skipped due to bug #565. Now that the bug is fixed, the `test.skip` has been removed to re-enable the test.

## Summary by CodeRabbit

* **Tests**
  * Re-enabled the "Delete profile" test using the action button so deletion is validated end-to-end.
  * Updated profile naming/creation logic in tests to use a new naming prefix for profiles with custom rules, improving test reliability.


abrugaro:
Closing as it's been already implemented.
Thanks!!

PR #1304, opened 2026-03-06
:bug: fix llm proxy race condition

djzager:
## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved startup race causing the local config to overwrite the Hub LLM proxy, and fixed stream errors caused by a leaked option.
  * Ensured TLS settings propagate correctly to proxy-backed models and added support for no-auth Hub setups.

* **Improvements**
  * Better provider selection to prefer the Hub proxy when available.
  * Enhanced diagnostics and logging for provider health and connection troubleshooting.


abrugaro (approved):
ack, I tested this build and I can't connect to the llm proxy yet but it may be for a different reason

PR #1327, opened 2026-03-25
:test_tube: llemulator for go accept/reject e2e workflow

hhpatel14:
## Summary by CodeRabbit

* **Tests**
  * Added deterministic emulator fixture and scripted responses to make Go E2E workflows repeatable and verifiable.
  * Promoted Go workflow tests to a higher tier and added provider-aware setup that conditionally loads and verifies emulator responses.
  * Added gating/skips for solution flows when no provider is available, increased "Accept" timeouts, and improved failure messages.
  * Improved stability with reloads before analysis, polling waits for re-analysis, and stronger acceptance/issue-count checks.

### DIFF
--- a/tests/e2e/fixtures/gotest-llemulator.ts
+++ b/tests/e2e/fixtures/gotest-llemulator.ts
@@ -1,5 +1,7 @@
 /**
  * Deterministic LLM responses for savitharaghunathan/gotest when using llemulator (CI).
+ * Loaded from `02-e2e-workflow.test.ts` when `getDefaultProviderConfig() === LLEMULATOR_PROVIDER`
+ * (same gating idea as `plugin-settings.test.ts`, with payloads extracted here for size).
  * Matches real file shapes so Get Solution → Accept can complete without a live model.
  *
  * The workflow applies two separate fixes (two Get Solution → Accept steps):
--- a/tests/e2e/fixtures/gotest-llemulator.ts
+++ b/tests/e2e/fixtures/gotest-llemulator.ts
@@ -12,9 +12,17 @@
  */
 import {
   buildKaiResponse,
+  getLlemulatorBaseUrl,
   loadLlemulatorResponses,
   type LlemulatorPatternRule,
 } from '../utilities/llemulator.utils';
+
+/** Matches only the first gotest rule (main.go), not the go.mod rule. */
+const MAIN_GO_PROBE_PROMPT =
+  'Migrate deprecated HorizontalPodAutoscaler from autoscaling/v2beta1 for nginx-hpa';
+
+/** Substring from MAIN_GO_MIGRATED — not present in GO_MOD_UPDATED. */
+export const GOTEST_MAIN_GO_RESPONSE_MARKER = 'Creating HPA using autoscaling/v2 API';
 
 /** main.go migrated from autoscaling/v2beta1 to autoscaling/v2 (addresses first two issues). */
 const MAIN_GO_MIGRATED = `package main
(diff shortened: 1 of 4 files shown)


tests/e2e/fixtures/gotest-llemulator.ts
```diff
@@ -0,0 +1,266 @@
...
+    reset: true,
+    responses: gotestLlemulatorRules(),
+  });
+}
+
+/**
+ * Confirms the main.go Kai script is active: a probe message matches the autoscaling rule and
+ * returns the migrated file body (not go.mod). Call after `loadGotestWorkflowLlemulatorResponses`.
+ */
+export async function verifyGotestMainGoLlemulatorRule(
```
abrugaro: I don't think we need this function at all. The llemulator utils already loads the operation required by the extension to verify the connection to the llm provider so if it fails the test won't pass.

tests/e2e/tests/golang/02-e2e-workflow.test.ts
```diff
@@ -256,37 +299,58 @@ test.describe.serial(
...
-        'Accept button not found. This may occur if the model is unreachable.'
-      ).toBeVisible({ timeout: 30000 });
+        'Accept button not found. This may occur if the model is unreachable, or CI has no OPENAI_API_KEY and TEST_LLEMULATOR_URL (scripted responses).'
+      ).toBeVisible({ timeout: 120000 });
       await acceptButton.click();
       console.log('Autoscaling fix accepted');
 
+      // Don't wait for notification — partial analysis fires and auto-dismisses
+      // before waitForAnalysisCompleted() starts watching (race condition).
       await vscodeApp.waitDefault();
```
abrugaro: This comment is not true at all, we should be able to replace the explicit timeout with:

```
    await vscodeApp.waitForAnalysisCompleted();
```

Even if the analysis completes instantly, the "analysis completed" notification will remain visible for a few secs

tests/e2e/tests/golang/02-e2e-workflow.test.ts
```diff
@@ -325,29 +389,45 @@ test.describe.serial(
...
       // Click Accept button
       const acceptButton = resolutionView.getByRole('button', { name: 'Accept' }).first();
-      await expect(acceptButton).toBeVisible({ timeout: 30000 });
+      await expect(acceptButton).toBeVisible({ timeout: 120000 });
       await acceptButton.click();
       console.log('Dependency fix accepted');
 
+      // Don't wait for notification — partial analysis fires and auto-dismisses
+      // before waitForAnalysisCompleted() starts watching (race condition).
+      await vscodeApp.waitDefault();
```
abrugaro: Same as https://github.com/konveyor/editor-extensions/pull/1327/changes#r3200729585

abrugaro (commented):
Nice bug catch!

Some minor considerations

tests/e2e/tests/golang/02-e2e-workflow.test.ts
hhpatel14: In CI run 25374875590, `waitForAnalysisCompleted()` timed out after 10 minutes  the 'Analysis completed successfully!' notification never appeared after Accept.  But i can put  `waitForAnalysisCompleted()`  before accept button is clicked.

abrugaro: > In CI run 25374875590, waitForAnalysisCompleted() timed out after 10 minutes the 'Analysis completed successfully!' notification never appeared after Accept.

That's weird, it may be a bug since the notification should appear and stay visible for seconds even if tha analysis completes instantly

>  But i can put waitForAnalysisCompleted() before accept button is clicked.

This won't work because the method will wait synchronously and the analysis won't be triggered until the solution is accepted

hhpatel14: The issue is partialAnalysis not being handle properly by go Analyzer. It works fine with full analysis but with partial analysis "Analysis completed successfully!" never shown. I think adding full analysis after each accept button should solve the problem for now. I have already open an issue for partial analysis for go analyzer. We will update the test when the issue is fixed.

abrugaro (approved):
acked

PR #1344, opened 2026-04-08
:bug: defer webview CSS loading to fix Firefox in DevSpaces

djzager:
In VS Code Web (Eclipse Che / DevSpaces), webview resources are served by a service worker that intercepts requests to synthetic hostnames. When a stale service worker is still controlling the page during a controller handoff, Firefox lets fetch requests fall through to real DNS resolution -- which fails because the hostname exceeds the 63-char DNS label limit and contains invalid characters (+). This causes the stylesheet to transfer 0 bytes, leaving the webview completely unstyled.

Replace the static <link> tag with a script that:
- Defers stylesheet injection until navigator.serviceWorker.ready resolves (when a service worker controller is present)
- Loads immediately in desktop VS Code (Electron) where no service worker is involved
- Retries up to 3 times on load failure with 1.5s delays

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated stylesheet handling: the app now injects the stylesheet dynamically and re-injects it when the active service worker controller changes to help maintain consistent styling across service-worker-driven loads. Removed the prior static stylesheet link.


djzager:
this didn't fix anything. Closing.

PR #1358, opened 2026-04-22
:test_tube: Get tier1 working in devspaces

### LINKED ISSUE #1356: Update collect-assets script
The generic provider has been removed in https://github.com/konveyor/analyzer-lsp/pull/1142 and now, the collect-assets script fails when trying to grab it from main. 

Update the script so it doesn't try to fetch the generic provider and ensure the analysis results remain the same

### LINKED ISSUE #1357: Get tier1 working in devspaces

abrugaro:
Resolves #1357 
Resolves #1356 

## Summary by CodeRabbit

* **Tests**
  * Improved e2e reliability: better workspace handling, clipboard support, longer timeouts, tighter synchronization, and more robust terminal command visibility handling.
* **Extension/Workspace**
  * Improved extension install/uninstall flow and workspace restart/recovery interactions for more reliable test runs.
* **Packaging**
  * Updated binary/dependency packaging and asset mapping for analyzer/providers.
* **Fixtures**
  * Updated static-report test fixture counts.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -73,7 +73,9 @@ jobs:
         run: npm ci
 
       - name: Download analyzer binary assets (for custom binary tests)
-        run: npm run collect-assets
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: npm run collect-assets -- --use-workflow-artifacts --branch=main
 
       - name: Upload downloaded assets (for custom binary tests)
         uses: actions/upload-artifact@v4
--- a/scripts/collect-assets.js
+++ b/scripts/collect-assets.js
@@ -276,16 +276,16 @@ const actions = [
     }),
   }),
 
-  // Extract generic-external-provider binaries to platform-specific directories (same as kai pattern)
+  // Extract nodejs-external-provider binaries to platform-specific directories (same as kai pattern)
   async () => ({
-    id: "generic-external-provider binaries",
+    id: "nodejs-external-provider binaries",
     meta: await unpackAssets({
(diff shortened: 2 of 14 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -73,7 +73,9 @@ jobs:
         run: npm ci
 
       - name: Download analyzer binary assets (for custom binary tests)
-        run: npm run collect-assets
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: npm run collect-assets -- --use-workflow-artifacts --branch=main
```
djzager, line 78: Might need to think about this use of `main` here. I'm concerned this will cause odd behavior on release branches.

djzager (changes requested):
Only one request, look more closely please at the use-workflow-artifacts to determine if we need to be more cautious for this to work well on release branches.

.github/workflows/e2e-tests.yml
abrugaro: If I'm correct, the [release workflow ](https://github.com/konveyor/editor-extensions/blob/main/.github/workflows/release.yml#L89) always picks the same hardcoded [release-tag](https://github.com/konveyor/editor-extensions/blob/main/scripts/collect-assets.js#L20), so, as you pointed out, we may be having odd behavior on new release branches that fetch the latest analyzer changes (and/or if we backport anything).

I've created [this task](https://github.com/konveyor/editor-extensions/issues/1359) to look further into that.

The changes I made do unblock the merges to main, but let me know if you want me to revert them until more research is done on how to approach this "versioning"

djzager (approved):
ACK

PR #1367, opened 2026-04-25
:bug: capture in-place user edits when sending accepted solution to solution server

### LINKED ISSUE #1362: User edits to solutions not sent to solution server
## Description

When a user receives a solution from the LLM, opens it in the editor, makes modifications, and then accepts it, the user's edits do not appear to be sent to the solution server. Only the accept/reject signal for the original LLM-proposed solution is communicated — the modified content is lost from the solution server's perspective.

Communication with the solution server is otherwise working normally.

## Status: Needs Verification

> **Note:** This was reported by a user and needs further investigation to confirm the exact behavior. It's possible the modifications are being sent but not processed, or that there's a gap in what the accept flow transmits.

1. Run analysis and receive a solution suggestion from the LLM
2. Open the solution in the editor
3. Make modifications to the proposed solution
4. Accept the solution
5. Observe that the solution server receives the accept but not the user's modifications

## Expected Behavior

When a user modifies a proposed solution before accepting, the modified version should be sent to the solution server, not just the accept/reject status of the original.

ibolton336:
Resolves https://github.com/konveyor/editor-extensions/issues/1362 
## Problem

When the LLM proposes a fix and the user edits the suggested file directly in the editor before clicking **Accept**, the solution server was receiving the **original LLM-generated content** rather than the user's modified version. This meant user edits were silently discarded from the solution server's learning loop — the solution server would never see the human-corrected version of the fix.

This was reported by a consultant who noted that in-place editing didn't result in modifications being sent to the solution server, so his edits wouldn't get turned into hints for future migrations.

## Root Cause

In `handleFileResponse.ts`, when notifying the solution server of an accepted change, the code was passing `fileContent` — the LLM-generated content captured at fix generation time and stored in the chat message state:

```typescript
// Before: sends stale LLM content, ignores user edits
await executeExtensionCommand("changeApplied", path, fileContent);
```

This content is never updated when the user makes edits to the file in the editor after the fix is generated.

## Fix

Read the file from disk immediately before notifying the solution server. This captures any edits the user made between fix generation and acceptance:

```typescript
// After: reads current on-disk content, capturing in-place edits
let finalContent = fileContent;
try {
  const diskBytes = await vscode.workspace.fs.readFile(uri);
  finalContent = new TextDecoder().decode(diskBytes);
  logger.info(`Using on-disk content for solution server (captures in-place edits): ${path}`);
} catch (readError) {
  logger.warn(`Could not read disk content, using original content: ${path}`, readError);
}
await executeExtensionCommand("changeApplied", path, finalContent);
```

The disk read falls back to the original LLM content if the file cannot be read, preserving existing behavior for error cases.

## Testing

Verified end-to-end locally:
1. Run analysis on coolstore
2. Get Solution for `jms-to-reactive-quarkus-00010`
3. Edit the suggested fix in the editor before clicking Accept
4. Click Accept
5. Extension log confirms: `Using on-disk content for solution server (captures in-place edits): .../OrderServiceMDB.java`
6. Solution server stores the user-edited version (`Solution N status: modified`)
7. Subsequent "Get Solution" on same violation returns the stored hint ✅

## Related

- Related: https://github.com/konveyor/kai/issues/923

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where user edits to generated solutions were not being captured when accepted. The system now properly reads the latest file content at acceptance time.

### DIFF
--- a/changes/unreleased/fix-solution-server-inline-edit-capture.yaml
+++ b/changes/unreleased/fix-solution-server-inline-edit-capture.yaml
@@ -1,0 +1,11 @@
+kind: bugfix
+
+description: >
+  Fix solution server not receiving user's in-place edits when accepting a solution.
+  Previously, when a user edited the LLM-suggested fix directly in the editor before
+  clicking Accept, the solution server received the original LLM-generated content
+  instead of the user's modified version. The fix now reads the file from disk at
+  accept time, capturing any edits made between fix generation and acceptance.
+
+extensions:
+  - core
--- a/vscode/core/src/utilities/ModifiedFiles/handleFileResponse.ts
+++ b/vscode/core/src/utilities/ModifiedFiles/handleFileResponse.ts
@@ -204,11 +204,20 @@ export async function handleFileResponse(
       }
 
       // Notify solution server of the change
+      // Read from disk to capture any in-place edits the user made before accepting
       try {
         if (isDeleted) {
           await executeExtensionCommand("changeDiscarded", path);
         } else {
-          await executeExtensionCommand("changeApplied", path, fileContent);
+          let finalContent = fileContent;
+          try {
+            const diskBytes = await vscode.workspace.fs.readFile(uri);
+            finalContent = new TextDecoder().decode(diskBytes);
+            logger.info(`Using on-disk content for solution server (captures in-place edits): ${path}`);
+          } catch (readError) {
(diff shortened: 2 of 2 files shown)


djzager:
This looks like a backport...remove it if I'm wrong.

djzager (approved):
lgtm

abrugaro (approved):

PR #1368, opened 2026-04-28
:sparkles: unify agent backends on ACP protocol and remove dead queue pipeline

djzager:
Extract AgentBackendClient interface and shared types into dedicated module. Move backend-specific defaults (binaryName, binaryArgs, minimumVersion) to config layer. Replace all `as any` casts in acpClient.ts with proper SDK type narrowing using discriminated unions.


ibolton336 (commented):
## Review: ✨ Unify agent backends on ACP protocol and remove dead queue pipeline

### Summary

Massive simplification: replaces two separate clients (`GooseClient` + `OpencodeAgentClient`) with a single `AcpClient` using `@agentclientprotocol/sdk`, removes the dead `MessageQueueManager` pipeline (~3,200 lines), and unifies config into a single `agentBinaryPath` setting.

---

### 👍 What's Good

- **Huge maintenance win** — removing the queue/interaction machinery (processMessage, queueManager, handleModifiedFile, handleQuickResponse, processModifiedFile) eliminates timeout bombs, `as any` casts, and complex promise-based user interaction flows.
- **ACP SDK adoption** — `ClientSideConnection` + `ndJsonStream` instead of hand-rolled JSON-RPC eliminates framing bugs.
- **Extensibility** — `BACKEND_CONFIGS` map with fallback means adding a third ACP agent is zero-code.
- **DirectLLMClient is self-contained** — translating workflow events in-place via `translateWorkflowEvent` makes the non-agent path much easier to reason about.
- **Type safety** — `AgentBackendClient` interface is cleaner; no more `getWorkflow?()` optional method.

---

### ⚠️ Concerns / Questions

**1. `as any` in DirectLLMClient event translation**
`const chunk = msg.data as any;` — since this is a clean-slate refactor, could we type `KaiWorkflowMessage.data` more precisely (even `Record<string, unknown>` would be safer)?

**2. Permission handler cast**
```ts
this.handlePermissionRequest(client as AcpClient, data);
```
Works because only `AcpClient` emits `permissionRequest`, but if a future bug makes `DirectLLMClient` emit it, the cast hides the type error. Consider guarding with `instanceof` or making the handler accept `AgentBackendClient`.

**3. `STOP_WORKFLOW` no longer cancels the agent**
The old handler stopped the workflow and cleaned up the queue. The new one just mutates state. If `AcpClient.sendMessage()` is still awaiting, nothing calls `cancelGeneration()`. Should this call it to stop the agent process from continuing to generate?

**4. MCP server config in `createAcpClient`**
The old code wired an `mcpServerEntry` for the Konveyor MCP bridge. That's gone from `createAcpClient`. Is it wired elsewhere (e.g., `init.ts`'s `setMcpServers`), or intentionally dropped for this branch?

**5. Breaking config change**
Renaming `gooseBinaryPath` → `agentBinaryPath` resets users' existing settings. No migration path. Fine for `feature/goose` branch, but worth calling out in release notes.

---

### 🔬 Nits

- `AcpClient.updateModelEnv` casts `this.config as AcpClientConfig` — unnecessary since it's already that type.
- `disposed` field in `AcpClient` is set but never read (unlike old `GooseClient` which checked it in stream handling).
- `createSession` duplicates the `mcpServers` mapping logic from `start()` — could extract to a helper.

---

### Verdict

Approve-worthy with minor suggestions. The deletion of the queue pipeline and consolidation onto ACP is the right direction. Remaining concerns are "nice to fix" (worst case is UX, not data loss). Clean PR for its scope.

PR #1369, opened 2026-04-28
:bug: clean up diff state when editor tab is closed (#1362)

### LINKED ISSUE #1362: User edits to solutions not sent to solution server

ibolton336:
## Summary

Fixes three related issues when closing an editor tab with active diff decorations (review mode):

1. **Continue button stuck disabled** — `activeDecorators` not cleaned up on tab close
2. **File shows patched content after close** — editor edits silently fail on closing document, leaving dirty buffer
3. **Can't re-open review** — `viewingInEditor` state not reset, UI stuck on "All changes resolved"

## Problem

When the user opens a solution in review mode (vertical diff with accept/reject code lenses) and closes the editor tab without explicitly accepting or rejecting:

- The `activeDecorators` state is not cleaned up → continue button stays disabled
- `handler.clear(false)` tries to run `editor.edit()` on a closing document → edits silently fail → VS Code's in-memory buffer retains the patched content (even though `git status` shows no changes)
- The webview's local `viewingInEditor` state is never reset → UI shows "All changes resolved" instead of returning to "📝 Review in Editor"

## Root Cause

`VerticalDiffManager` had no `onDidCloseTextDocument` listener. It only cleaned up diff state on:
- Explicit accept/reject actions via code lenses
- Active editor switches (`onDidChangeActiveTextEditor`)

Closing an editor tab triggers neither of these.

### 1. `onDidCloseTextDocument` listener (`manager.ts`)
Added a listener that detects when a document with an active diff handler is closed.

### 2. `clearForClosedDocument` method (`manager.ts`)
New method specifically for tab-close cleanup that does NOT attempt editor edits (which fail on closing documents). Instead it:
- Clears all internal state maps (handler, codelens, streamId, activeDecorators)
- Disposes the handler without running accept/reject edits
- Reverts the file via `workbench.action.files.revert` so VS Code drops the dirty buffer and reloads clean content from disk

### 3. Reset webview review state (`BatchReviewExpandable.tsx`)
Added a `useEffect` that watches `activeDecorators` and clears the local `viewingInEditor` state when the decorator for the currently viewed file disappears. This drops the UI back to the default state showing "📝 Review in Editor", allowing the user to re-open the review.

## Testing

1. Run analysis and receive a solution suggestion
2. Open the solution in review mode (click to see code lenses)
3. Close the editor tab (X button) without accepting or rejecting
4. Verify:
   - ✅ The continue button in the batch review is re-enabled
   - ✅ The file shows original (clean) content when re-opened
   - ✅ The "📝 Review in Editor" button is shown again (not stuck on "All changes resolved")
   - ✅ Clicking "📝 Review in Editor" re-opens the diff review successfully

Fixes: https://github.com/konveyor/editor-extensions/issues/1362
Jira: [MTA-6862](https://redhat.atlassian.net/browse/MTA-6862), [MTA-6863](https://redhat.atlassian.net/browse/MTA-6863)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed review mode issue where closing an editor tab without acting on accept/reject code lenses would disable the batch review continue button. The review UI now properly resets and files revert to their on-disk state when tabs are closed.

### DIFF
--- a/changes/unreleased/fix-review-mode-continue-button-blocked.yaml
+++ b/changes/unreleased/fix-review-mode-continue-button-blocked.yaml
@@ -6,7 +6,9 @@ description: >
   the editor tab without accepting or rejecting, the activeDecorators state was not cleaned
   up. This left the batch review UI in a stuck state where the continue button stayed disabled.
   Added an onDidCloseTextDocument listener to VerticalDiffManager that clears diff state when
-  the editor tab is closed, treating it as a reject/discard.
+  the editor tab is closed, treating it as a reject/discard. Also fixes a related issue where
+  the closed file retained patched content in VS Code's in-memory buffer by reverting the file
+  to its on-disk state, and resets the webview review UI so users can re-open the review.
 
 extensions:
   - core
--- a/changes/unreleased/fix-review-mode-continue-button-blocked.yaml
+++ b/changes/unreleased/fix-review-mode-continue-button-blocked.yaml
@@ -1,0 +1,12 @@
+kind: bugfix
+
+description: >
+  Fix continue button remaining disabled after closing editor tab with active code lenses.
+  When the user opens a solution in review mode (with accept/reject code lenses) and closes
+  the editor tab without accepting or rejecting, the activeDecorators state was not cleaned
+  up. This left the batch review UI in a stuck state where the continue button stayed disabled.
+  Added an onDidCloseTextDocument listener to VerticalDiffManager that clears diff state when
+  the editor tab is closed, treating it as a reject/discard.
+
+extensions:
+  - core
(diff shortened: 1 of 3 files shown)


djzager (approved):
lgtm

PR #1372, opened 2026-04-28
:bug: retry java activation

### LINKED ISSUE #1349: Java extension should retry activation when Red Hat Java extension is not yet ready
### Problem

When loading in DevSpaces (or similar remote environments), the Konveyor Java extension and the Red Hat Java extension activate concurrently. The Red Hat Java extension often takes additional time to reach Standard mode. The Konveyor Java extension currently makes a **single, synchronous check** at activation time ([extension.ts:72-100](https://github.com/konveyor/editor-extensions/blob/main/vscode/java/src/extension.ts#L72-L100)):

1. `vscode.extensions.getExtension("redhat.java")` — if not found, abort
2. `await javaExt.activate()` — single attempt, abort on failure

If the Red Hat Java extension isn't ready at that exact moment, the Konveyor Java extension fails to activate and never recovers. The user must reload the browser/window to trigger re-activation.

### Current behavior

- Extension activation is one-shot: no retry, no polling, no event listeners
- Health checks (`healthCheck.ts`) detect the problem after the fact but can't trigger re-activation
- User must manually reload to recover

### Expected behavior

The extension should gracefully handle the case where `redhat.java` is not yet ready at activation time, and automatically complete initialization once it becomes available — without requiring a window reload.

### Essential minimum

Use **`vscode.extensions.onDidChange`** to defer provider registration when `redhat.java` is not active at startup:

1. If `redhat.java` is not found or not active during `activate()`, register a listener on `vscode.extensions.onDidChange`
2. On each change event, re-check `vscode.extensions.getExtension("redhat.java")` and whether it's active
3. Once active, proceed with the rest of initialization (LSP proxy, provider manager, core registration)
4. Add a reasonable timeout (e.g., 2-3 minutes) after which a notification is shown with a "Reload Window" action
5. Don't block — `activate()` should return normally, and the deferred init should run in the background

This is the simplest approach that avoids polling/timers and uses VS Code's built-in event. It does **not** require checking `java.server.mode` or any Red Hat Java internals.

### Nice-to-have (opinions, not requirements)

- **Show a progress notification** while waiting (e.g., "Waiting for Java Language Server...") so users know something is happening, not broken
- **Check `java.server.mode`** — the Red Hat Java extension can report `isActive: true` while still in Lightweight mode. Checking mode would be more precise, but historically these APIs have been unreliable and fragile to depend on. If the Red Hat Java extension exposes a stable readiness signal, we could use it, but `isActive` is likely sufficient for the race condition described here
- **Retry the full provider startup** — if `java-external-provider` or the LSP proxy fails because the language server wasn't ready, those could also benefit from retry logic. But that's a separate, larger concern

### Context

- Reported in the context of DevSpaces workshops where concurrent extension loading causes a race condition
- The older monolithic extension may have had `java.server.mode` checks but they were removed due to unreliability
- The current multi-extension architecture (core + java) was introduced in #960

djzager:
Fixes #1349 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race condition that caused the Java extension to fail during activation and require a window reload. The extension now waits for the Java Language Server to initialize before starting, with a 3-minute timeout safeguard to alert users if initialization takes too long.

### DIFF
--- a/changes/unreleased/1349-java-retry-activation.yaml
+++ b/changes/unreleased/1349-java-retry-activation.yaml
@@ -1,0 +1,8 @@
+kind: bugfix
+description: >
+  Java extension now waits for the Java Language Server to reach Standard mode
+  before starting the LSP proxy and provider, fixing a race condition in
+  DevSpaces where the extension would fail to activate and require a window
+  reload.
+extensions:
+  - java
--- a/vscode/java/src/extension.ts
+++ b/vscode/java/src/extension.ts
@@ -15,6 +15,18 @@ import { execFile } from "child_process";
 import { promisify } from "util";
 import { javaHealthChecks } from "./healthCheck";
 
+/** How long to wait for JDTLS Standard mode before giving up */
+const JDTLS_READY_TIMEOUT_MS = 3 * 60 * 1000; // 3 minutes
+
+/**
+ * API shape exported by the Red Hat Java extension (redhat.java).
+ * Only the subset we depend on.
+ */
+interface RedHatJavaExtensionApi {
+  serverReady(): Promise<boolean>;
+  onDidServerModeChange: vscode.Event<string>;
+}
+
 /**
  * Check if a command is available on the system
  */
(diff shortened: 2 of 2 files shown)


djzager:
```
{"level":"info","message":"Logger created","timestamp":"2026-04-28T18:47:17.805Z"}
{"level":"info","message":"Extension konveyor.konveyor-java starting","timestamp":"2026-04-28T18:47:17.805Z"}
{"level":"info","message":"Waiting for Java Language Server to be ready (non-blocking)...","timestamp":"2026-04-28T18:47:17.805Z"}
{"level":"info","message":"Java Language Server mode changed to: Standard","timestamp":"2026-04-28T18:47:25.697Z"}
{"level":"info","message":"Java installation detected","timestamp":"2026-04-28T18:47:25.976Z"}
{"level":"info","message":"Maven installation detected","timestamp":"2026-04-28T18:47:26.450Z"}
{"level":"info","message":"Found core extension, activating...","timestamp":"2026-04-28T18:47:26.451Z"}
...
```

abrugaro (approved):
acked

PR #1374, opened 2026-04-29
:bug: Add enabled property to llm proxy client

### LINKED ISSUE #1365: LLM proxy attempts connection even when disabled by operator
## Description

When the LLM proxy is disabled by the operator in a DevSpaces environment, the extension still attempts to connect to the proxy. This causes unnecessary connection attempts and potentially confusing error messages or behavior.

## Status: Needs Verification

> **Note:** This was reported by a user in a DevSpaces environment and needs investigation to confirm whether the extension is not reading the operator configuration correctly, or if there is a race condition / caching issue with the disabled state.

1. Set up a DevSpaces environment where the operator has disabled the LLM proxy
2. Open the extension
3. Observe that the extension still attempts to connect to the LLM proxy

## Expected Behavior

When the LLM proxy is disabled by the operator, the extension should not attempt to connect to it.

abrugaro:
Resolves #1365 

Operator PR with new configmap property: https://github.com/konveyor/operator/pull/558

If the `enabled` property is not present in the configmap, it is assumed to be true to ensure compatibility with older versions of the operator. To prevent the bug from occurring, both the operator and the extension will need to be updated

## Summary by CodeRabbit

* **Bug Fixes**
  * The LLM proxy client now supports an `enabled` configuration setting to control whether the proxy is active in your environment.

### DIFF
--- a/changes/unreleased/1374-add-enavled-property-to-llmproxy-client.yaml
+++ b/changes/unreleased/1374-add-enavled-property-to-llmproxy-client.yaml
@@ -1,3 +1,3 @@
 kind: bugfix
 description: >
-  TODO: Describe your change here.
+  Add enabled property to LLM proxy client.
--- a/changes/unreleased/1374-add-enavled-property-to-llmproxy-client.yaml
+++ b/changes/unreleased/1374-add-enavled-property-to-llmproxy-client.yaml
@@ -1,0 +1,3 @@
+kind: bugfix
+description: >
+  TODO: Describe your change here.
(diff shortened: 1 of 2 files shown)


djzager:
Adding release-0.4 cherry-pick label. Remove if I'm wrong please.

changes/unreleased/1374-add-enavled-property-to-llmproxy-client.yaml
```diff
@@ -0,0 +1,3 @@
+kind: bugfix
+description: >
+  TODO: Describe your change here.
```
djzager: A bit more descriptive 😎

abrugaro: Added, now with 100% more files saved...

djzager (approved):

PR #1392, opened 2026-05-01
:bug: fix: add configurable LLM request timeout to prevent hanging

ibolton336:
## Summary

Adds a configurable timeout (default 5 minutes) to all LLM requests in `streamOrInvoke`. Previously, a misconfigured or unreachable model provider would hang `streamOrInvoke` indefinitely with no error.

### Changes

- `agentic/src/nodes/base.ts` — wraps LLM execution in `Promise.race` with a timeout
  - Default: 5 minutes (`DEFAULT_LLM_TIMEOUT_MS = 300_000`)
  - Configurable per-call via `timeoutMs` option in `streamOptions`
  - Clear error message: "Verify your provider-settings.yaml configuration is correct"
  - Properly cleans up timer and suppresses unhandled rejection from the racing promise
  - Error workflow message emission gated on `emitResponseChunks` flag (prevents internal "thinking" operations from leaking errors to the user-facing stream)

### Testing
- No new tests (existing tests unaffected — timeout only fires on actual provider hangs)
- Verified locally with misconfigured provider ✅

### DIFF
--- a/agentic/src/nodes/analysisIssueFix.ts
+++ b/agentic/src/nodes/analysisIssueFix.ts
@@ -325,7 +325,10 @@ If you have any additional details or steps that need to be performed, put it he
     );
 
     if (!response) {
-      this.logger.silly("AnalysisIssueFix returned undefined response");
+      this.logger.warn(
+        `AnalysisIssueFix: LLM returned no response for file "${fileName}". ` +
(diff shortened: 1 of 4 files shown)


djzager (approved):
ACK

PR #1412, opened 2026-05-14
:bug: restore clean buffer on tab close via tabGroups.onDidChangeTabs (#1362)

ibolton336:
## Summary

Fixes the remaining issue from #1362 where closing a tab with active diff decorations (review mode) leaves the file showing patched content — even though `git status` shows clean. The only workaround was reopening VS Code.

Additionally fixes a pre-existing bug where the codelens "Accept All Changes" / "Reject All Changes" buttons silently failed and never advanced the batch review.

### 1. Dirty buffer on tab close
The previous fix (PR #1369) used `onDidCloseTextDocument` + `workbench.action.files.revert`. However, `onDidCloseTextDocument` fires **after** the document is detached from VS Code's editor model — `revert` has no active editor to target and silently fails. The dirty buffer persists in VS Code's hot-exit cache.

### 2. Accept/Reject All codelens not working (path vs URI mismatch)
The codelens provider passed `document.uri.fsPath` (a filesystem path like `/path/to/file.ts`) to the `acceptDiff`/`rejectDiff` commands, but `acceptAll`/`rejectAll` look up handlers by URI string (`file:///path/to/file.ts`). The mismatch caused handler lookups to always fail silently.

### 3. Batch review not advancing after codelens accept/reject
Even after fixing the URI mismatch, the codelens accept/reject cleared `activeDecorators` (sending `DECORATORS_UPDATE` to the webview) before removing the file from `pendingBatchReview`. The webview's `useEffect` reacted to the decorator clearing by resetting `viewingInEditor` to null, causing the "Accept/Reject/Review in Editor" buttons to flash back before the batch removal arrived.

### Tab close cleanup
Replace `onDidCloseTextDocument` with `tabGroups.onDidChangeTabs` (VS Code 1.77+), which fires while the document is still in the model. On tab close:

1. Dispose the diff handler **without** running editor edits (they'd silently fail)
2. Clear all internal state (activeDecorators, codelens, streamId)
3. Read clean on-disk content via `workspace.fs.readFile`
4. Replace the document buffer via `WorkspaceEdit` — content matches disk → `isDirty` = false → hot-exit won't preserve the patched state

### Codelens URI fix
Pass `fileUri` (from `document.uri.toString()`) instead of `document.uri.fsPath` for the Accept All / Reject All codelens arguments, and update the command handlers consistently.

### Batch review advancement
- Remove `activeDecorators` clearing from `clearForFileUri` and `onStatusUpdate("closed")` — only `clearForClosedTab` clears it now (for the tab-close → "Review in Editor" re-show behavior)
- `acceptDiff`/`rejectDiff` commands capture the `messageToken` before clearing, then call `handleFileResponse` + remove from `pendingBatchReview` + clear decorator (in that order)
- Added `onAllBlocksResolved` callback on the manager for the individual block-by-block case

This ensures the file is removed from `pendingBatchReview` **before** the decorator is cleared, so the webview never re-renders buttons for a file that's already gone.

### Tab close scenario
1. Run analysis, receive a solution
2. Click "Review in Editor" to open the vertical diff
3. Close the editor tab (X) without accepting or rejecting
4. Reopen the file — should show original clean content (no patched lines)

### Accept/Reject All scenario
1. Run analysis, receive a solution with diff blocks
2. Click "Review in Editor" to open the vertical diff
3. Click "✓ Accept All Changes" or "✗ Reject All Changes" at the top
4. Changes should apply/revert, codelens disappears, batch review advances to next file

### Individual block scenario
1. Open diff with multiple blocks
2. Accept/reject blocks one by one via per-block codelens
3. After last block is resolved, batch review should advance

Refs: #1362

---
`cherry-pick`

## Summary by CodeRabbit

* **Bug Fixes**
  * Restores documents to their on-disk state when closing a review editor tab without accepting changes.

* **Improvements**
  * More robust batch-review flow to avoid race conditions when accepting or rejecting changes.
  * Automatically saves resolved files and shows success notifications when changes are applied or rejected.

* **Refactor**
  * Tab-close and diff-resolution handling updated for more reliable behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/konveyor/editor-extensions/pull/1412)

### DIFF
--- a/changes/unreleased/1410-fix-dirty-buffer-on-tab-close.yaml
+++ b/changes/unreleased/1410-fix-dirty-buffer-on-tab-close.yaml
@@ -1,0 +1,6 @@
+kind: bugfix
+description: >
+  Fix file showing patched content after closing review editor tab without
+  accepting. Use tabGroups.onDidChangeTabs to restore the clean on-disk buffer
+  via WorkspaceEdit instead of the unreliable onDidCloseTextDocument + revert
+  approach.
--- a/vscode/core/src/commands.ts
+++ b/vscode/core/src/commands.ts
@@ -37,6 +37,7 @@ import { fixGroupOfIncidents, IncidentTypeItem } from "./issueView";
 import { paths } from "./paths";
 import { checkIfExecutable, copySampleProviderSettings } from "./utilities/fileUtils";
 import { handleConfigureCustomRules } from "./utilities/profiles/profileActions";
+import { handleFileResponse } from "./utilities/ModifiedFiles/handleFileResponse";
 import { VerticalDiffCodeLensProvider } from "./diff/verticalDiffCodeLens";
 import type { Logger } from "winston";
 import { parseModelConfig, getProviderConfigKeys } from "./modelProvider/config";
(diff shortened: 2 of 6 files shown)


abrugaro (approved):
Tested with the latest ci build.
lgtm

PR #1431, opened 2026-06-02
:bug: honor NO_PROXY when selecting proxy dispatcher

fabianvf:
Closes #1415.

When `HTTP_PROXY` / `HTTPS_PROXY` are set in the shell that launches VS Code, the extension was routing every Hub/provider connection through that proxy — even when `NO_PROXY` listed the target host. So a user with a corporate proxy in their environment and a local model on `http://127.0.0.1:8080/v1` would see:

```
Using proxy for Hub/provider connections: http://172.19.44.13:5555/
```

even though their `NO_PROXY=127.0.0.1` should have bypassed it. The only workaround was unsetting `HTTP_PROXY`/`HTTPS_PROXY` before launching VS Code.

### Root cause

`getDispatcherWithCertBundle` and `getNodeHttpHandler` in `vscode/core/src/utilities/tls.ts` read `HTTP_PROXY`/`HTTPS_PROXY` and unconditionally constructed a `ProxyAgent` / `HttpsProxyAgent`. Neither function accepted a target URL, so `NO_PROXY` couldn't be evaluated.

### Fix

- New `shouldBypassProxy(targetUrl, noProxy)` helper with curl/requests-compatible semantics: `*` wildcard, exact host, leading-dot suffix (`.example.com`), bare-host suffix, port-qualified entries (`localhost:8080`), case-insensitive, comma- or whitespace-separated.
- Both dispatcher factories now take an optional `targetUrl` and skip the proxy branch when `shouldBypassProxy` returns true. Old callers that don't pass a URL get the existing behavior — no surprise regressions.
- Wired the target URL through every call site:
  - `HubConnectionManager` — passes `this.config.url`
  - `modelCreator` — new `extractProviderTargetUrl(args)` helper pulls baseURL/baseUrl/endpoint/azureOpenAIEndpoint from the model args. For `ChatBedrock`, falls back to synthesizing `https://bedrock-runtime.{region}.amazonaws.com` from `AWS_DEFAULT_REGION` when no explicit endpoint is configured.
- Added a `"Bypassing proxy for X (matches NO_PROXY=Y)"` info log so the decision is debuggable from logs the way the original symptom was.

Also a small drive-by: the `Using proxy ... for AWS Bedrock` log line was leaking proxy basic-auth creds; now passed through `sanitizeUrl`.

### Tests

Built TDD-style — tests written first, watched them fail, then implemented.

- 17 unit cases for `shouldBypassProxy` covering each matching rule and a few false-positive guards (`badexample.com` mustn't match `example.com`, unparseable URLs return false instead of throwing, etc.)
- 5 integration cases for `getDispatcherWithCertBundle` and 2 for `getNodeHttpHandler` that set proxy env vars and assert the right dispatcher subclass is constructed — including the literal #1415 repro.
- All 132 tests pass (`npm test:unit-tests`).

### Test plan

- [x] Unit + integration tests in `vscode/core/src/utilities/__tests__/tls.test.ts`
- [ ] Manual: launch VS Code with `HTTPS_PROXY=http://fake-proxy:8080 NO_PROXY=127.0.0.1`, point a `ChatOpenAI` provider at `http://127.0.0.1:8080/v1`, confirm the `Using proxy for Hub/provider connections` log line no longer fires for that provider (and that the new `Bypassing proxy` line does)

### Known limitations

- `shouldBypassProxy` doesn't support CIDR ranges (`192.168.0.0/16`) or `*.example.com` glob syntax. Consistent with curl/requests/proxy-from-env; can be added later if there's demand.
- `ChatBedrock` endpoint synthesis assumes standard regional partition. FIPS / GovCloud / China-partition users will need to set `args.endpointUrl` explicitly (and probably already do).

## Summary by CodeRabbit

* **Bug Fixes**
  * Proxy handling now honors the NO_PROXY environment variable across Hub and model-provider connections; targets matching NO_PROXY (including localhost/127.0.0.1) will bypass proxies for more reliable local and self‑hosted connectivity.

* **Tests**
  * Added regression tests covering NO_PROXY matching semantics and proxy bypass behavior to prevent regressions.

### DIFF
--- a/changes/unreleased/1415-honor-no-proxy.yaml
+++ b/changes/unreleased/1415-honor-no-proxy.yaml
@@ -1,0 +1,6 @@
+kind: bugfix
+description: >
+  Honor the NO_PROXY environment variable when deciding whether to route
+  Hub and GenAI provider connections through HTTP_PROXY/HTTPS_PROXY.
+  Targets matching NO_PROXY (including loopback addresses such as
+  127.0.0.1) now bypass the proxy as expected.
--- a/vscode/core/src/hub/HubConnectionManager.ts
+++ b/vscode/core/src/hub/HubConnectionManager.ts
@@ -359,7 +359,13 @@ export class HubConnectionManager {
     // Create connection-scoped fetch for insecure mode
     if (this.config.auth.insecure) {
       this.logger.warn("SSL certificate verification is disabled for Hub connections");
-      const dispatcher = await getDispatcherWithCertBundle(undefined, true);
+      const dispatcher = await getDispatcherWithCertBundle(
+        undefined,
+        true,
+        false,
+        this.logger,
+        this.config.url,
+      );
       this.scopedFetch = getFetchWithDispatcher(dispatcher);
     } else {
       this.scopedFetch = null;
--- a/vscode/core/src/modelProvider/modelCreator.ts
+++ b/vscode/core/src/modelProvider/modelCreator.ts
@@ -33,7 +33,7 @@ class AzureChatOpenAICreator implements ModelCreator {
   constructor(private readonly logger: Logger) {}
 
   async create(args: Record<string, any>, env: Record<string, string>): Promise<BaseChatModel> {
-    const fetchFn = await setupProviderTLS(env, this.logger);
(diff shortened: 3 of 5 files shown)


djzager (approved):
LGTM and verified to address the issue. Just needs the DCO.

PR #1440, opened 2026-06-11
:sparkles: implement provider-specific rule packaging

savitharaghunathan:
Move from bundling all rulesets in core to each language extension supplying its own stable rulesets. Per-language extraction in collect-assets.js, per-language packaging in copy-dist.js, provider rulesets aggregation in analyzerClient, and re-discovery of labels on provider registration.

Refs: #1426

Assisted-by: Claude Code (claude-opus-4-6)

## Summary by CodeRabbit

* **New Features**
  * Added an event that notifies when a provider is registered.
* **Improvements**
  * Stable rulesets are now bundled by each language extension; preview rulesets are also included in the packaged output.
  * Analyzer configuration can aggregate rulesets from all registered providers (when default rules are enabled).
  * Label discovery now scans multiple ruleset directories and refreshes when new providers contribute rulesets.
* **Tests**
  * Added/updated coverage for active profile ruleset aggregation and multi-directory label discovery.

### DIFF
--- a/scripts/collect-assets.js
+++ b/scripts/collect-assets.js
@@ -150,9 +150,8 @@ const actions = [
   }),
 
   // Download and extract seed rulesets from a rulesets repo release.
-  // Supports both the layout introduced in konveyor/rulesets#342 (stable/<lang>/* and preview/*)
-  // and the legacy layout (default/generated/*). Tries the new layout first; falls back to the
-  // legacy one. Throws if neither layout produces any files.
+  // Per-language rulesets are extracted into separate directories for each language extension.
+  // Preview rulesets and legacy fallback go to the core extension's rulesets directory.
   async () => {
     const rulesetConfig = {
       downloadDirectory: join(DOWNLOAD_CACHE, "sources"),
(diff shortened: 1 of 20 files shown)


scripts/collect-assets.js
```diff
@@ -164,39 +163,63 @@ const actions = [
...
+    console.warn(
+      "No per-language rulesets layout found, falling back to legacy layout (default/generated)",
+    );
+    const meta = await downloadAndExtractGitHubReleaseSourceCode({
+      ...rulesetConfig,
+      context: "{{root}}/default/generated",
+    });
+    console.log("Extracted rulesets using legacy layout (default/generated)");
+    return { id: "seed rulesets", meta };
   },
```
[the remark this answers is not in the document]
savitharaghunathan: Not every language has upstream rulesets in the stable/<lang>/ layout today. For example, Go currently
  has no stable rulesets in the upstream release. Requiring all languages to have rulesets before
  accepting the per-language layout would break the entire feature whenever a new language extension is
  added before upstream rulesets exist for it.

  The current behavior:
  - Languages with rulesets get them extracted into rulesets-<lang>/
  - Languages without rulesets log a warning and get an empty directory — their extension simply
  contributes no default rules
  - The legacy fallback only kicks in when no per-language layout exists at all (i.e., an older upstream
  release that hasn't adopted the stable/<lang>/ structure)

  Forcing a fallback to the legacy monolithic bundle when one language is missing would defeat the purpose
  of the feature — every other language's rulesets would be bundled into core instead of their respective
  extensions.

@coderabbitai create an issue for this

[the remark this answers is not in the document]
savitharaghunathan: @djzager do you think the above approach makes sense?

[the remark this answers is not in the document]
djzager: Makes sense to me. Partial extraction is the right default — requiring all languages to succeed would be overly strict and break whenever a new language is added ahead of its upstream rulesets. The warning logs during collection are sufficient for diagnosing missing language rulesets.

vscode/core/src/client/__tests__/analyzerClient.test.ts
```diff
@@ -0,0 +1,130 @@
+import expect from "expect";
+
+// Test the getActiveProfileConfig logic in isolation.
+// AnalyzerClient has heavy dependencies (vscode, ChildProcess, etc.),
+// so we reproduce the exact aggregation logic in a testable wrapper.
+class TestableAnalyzerClient {
```
djzager: This test duplicates the `getActiveProfileConfig` logic in a standalone class rather than testing the real `AnalyzerClient`. If the real implementation changes, these tests still pass against stale code. Consider extracting the ruleset aggregation logic into a pure function (e.g., `buildRulesetsList(profile, coreRulesetsPath, providerRulesets)`) that both `AnalyzerClient` and this test can use, so the test exercises the actual code path.

vscode/java/src/extension.ts
```diff
@@ -356,8 +356,10 @@ async function initializeProviders(
       contextLines: 10,
     },
     rulesetsPaths: [
-      // In Phase 1, rulesets are still in core extension
-      // Will be moved to java extension in later phase
+      context.asAbsolutePath(
```
djzager: The `??` fallback is the same string as the dev-time value in `package.json`. It can only fire if `includedAssetPaths.rulesets` is somehow missing from the package manifest, which would indicate a packaging bug. Consider reading the value from `package.json` without a fallback and throwing an error if it's missing — that surfaces the packaging bug immediately instead of silently using a path that may not be correct in a packaged extension.

djzager (changes requested):
Nice work moving rulesets to per-language packaging. Two comments below.

vscode/core/src/client/__tests__/analyzerClient.test.ts
savitharaghunathan: @djzager thank you for the review :) I split this into a func

vscode/java/src/extension.ts
savitharaghunathan: addressed. ptal

djzager (approved):
Comments addressed — extracted buildRulesetsList and getIncludedRulesetsPath. Clean.

PR #1461, opened 2026-06-19
:bug: Surface analyzer error during server startup

abrugaro:
Resolves #1422 

Surface analyzer log errors to core extension's output during server status.

When the analyzer encounters errors during startup, these errors are only written to the analyzer.log file and never surfaced to the user. The extension's start button remains in a spinner state for 5 minutes with no indication of what went wrong, forcing users to manually locate and inspect the log file.

This:

```typescript
analyzerRpcServer.stderr.on("data", (data) => {
      // Log stderr output for debugging
      this.logger.debug(`Analyzer stderr: ${data.toString()}`);
```

Did not surface analyzer errors because they are written to the log file, not to stderr.

### Before

image

### After

image

## Summary by CodeRabbit

* **New Features**
  * Analyzer **ERROR/FATAL** messages are now streamed to the VS Code output channel while the analyzer starts.
  * Added live connection feedback while the extension waits for analyzer pipes, including early abort when startup fails and periodic “waiting” updates.
* **Bug Fixes**
  * Improved analyzer stderr visibility by logging most stderr output at **warning** level (with stage-related lines at lower verbosity).
  * Retry behavior now detects analyzer exit during wait periods and reports successful connections more clearly.

### DIFF
--- a/vscode/core/src/client/analyzerClient.ts
+++ b/vscode/core/src/client/analyzerClient.ts
@@ -370,15 +370,17 @@ export class AnalyzerClient {
     });
 
     analyzerRpcServer.stderr.on("data", (data) => {
-      const msg = data.toString().trimEnd();
-      if (!msg) {
-        return;
-      }
-      // Progress JSON (from -progress-output stderr) stays at debug level
-      if (msg.includes('"stage"')) {
-        this.logger.debug(`[analyzer-rpc-server stderr] ${msg}`);
-      } else {
-        this.logger.warn(`[analyzer-rpc-server stderr] ${msg}`);
+      for (const line of data.toString().split("\n")) {
+        const trimmed = line.trim();
+        if (!trimmed) {
+          continue;
+        }
+        // Progress JSON (from -progress-output stderr) stays at debug level
+        if (trimmed.includes('"stage"')) {
+          this.logger.debug(`[analyzer-rpc-server stderr] ${trimmed}`);
+        } else {
+          this.logger.warn(`[analyzer-rpc-server stderr] ${trimmed}`);
+        }
       }
     });
 
--- a/vscode/core/src/client/analyzerClient.ts
+++ b/vscode/core/src/client/analyzerClient.ts
@@ -32,6 +32,7 @@ import { TaskManager } from "src/taskManager/types";
 import { Logger } from "winston";
 import { executeExtensionCommand } from "../commands";
 import { ProviderRegistry } from "../api";
+import { AnalyzerLogTailer } from "./analyzerLogTailer";
 
 export class AnalyzerClient {
   // Progress percentage ranges for each stage
@@ -46,6 +47,7 @@ export class AnalyzerClient {
   private analyzerRpcServer: ChildProcessWithoutNullStreams | null = null;
   private analyzerRpcConnection?: rpc.MessageConnection | null;
   private currentProgressCallback?: (event: ProgressEvent) => void;
+  private logTailer: AnalyzerLogTailer | null = null;
 
   constructor(
     private extContext: vscode.ExtensionContext,
@@ -151,6 +153,7 @@ export class AnalyzerClient {
     const [analyzerRpcServer, analyzerPid] = this.startAnalysisServer(pipeName);
     analyzerRpcServer.on("exit", (code, signal) => {
       this.logger.info(`Analyzer RPC server terminated [signal: ${signal}, code: ${code}]`);
+      this.stopLogTailer();
       if (code) {
         vscode.window.showErrorMessage(
           `Analyzer RPC server failed. Status code: ${code}. Please see the output channel for details.`,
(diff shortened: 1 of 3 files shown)


vscode/core/src/client/analyzerLogTailer.ts
```diff
@@ -0,0 +1,94 @@
...
+    } catch {
+      return;
+    }
+
+    try {
+      const bytesToRead = currentSize - this.lastSize;
+      const buf = Buffer.alloc(bytesToRead);
+      fs.readSync(fd, buf, 0, bytesToRead, this.lastSize);
+      this.lastSize = currentSize;
+      this.processChunk(buf.toString("utf-8"));
```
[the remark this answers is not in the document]
abrugaro: I beileve the comment makes sense, but since the tailer should live for a short period (seconds) during the server startup, this should not be an issue.

vscode/core/src/client/analyzerLogTailer.ts
```diff
@@ -0,0 +1,94 @@
...
+    private readonly pollIntervalMs = 1000,
+  ) {}
+
+  start(): void {
+    if (this.intervalHandle) {
+      return;
+    }
+
+    this.lastSize = this.getCurrentFileSize();
+    this.buffer = "";
```
djzager, line 35: `lastSize` is initialized to the current file size, so errors written between process spawn and `start()` are skipped. Since this is a startup-only tailer and the analyzer creates a fresh log each run, initializing `lastSize = 0` would catch errors from the entire startup window — including the fast-crash scenario this feature is designed for.

vscode/core/src/client/analyzerClient.ts
```diff
@@ -342,11 +369,11 @@ export class AnalyzerClient {
...
 
-    // Progress updates are now received via RPC notifications (analysis.progress)
-    // The stderr-based progress parsing has been removed to avoid duplicate progress updates
     analyzerRpcServer.stderr.on("data", (data) => {
-      // Log stderr output for debugging
-      this.logger.debug(`Analyzer stderr: ${data.toString()}`);
+      const msg = data.toString().trimEnd();
+      if (msg) {
+        this.logger.warn(`[analyzer-rpc-server stderr] ${msg}`);
+      }
```
djzager, line 385: The analyzer is launched with `-progress-output stderr`, so progress JSON goes to stderr. Promoting all stderr from `debug` to `warn` means every progress update during analysis appears as a warning, flooding the output channel. Consider keeping progress JSON at `debug` and only promoting non-JSON/error-looking messages to `warn`, or leaving this at `debug` since the new log tailer already surfaces actual errors from `analyzer.log`.

vscode/core/src/client/analyzerClient.ts
```diff
@@ -253,9 +264,18 @@ export class AnalyzerClient {
...
   protected async getSocket(pipeName: string): Promise<Socket> {
-    const MAX_RETRIES = 150; // retry for 5 minutes to connect to the analyzer pipe
+    const MAX_RETRIES = 150;
     const RETRY_DELAY = 2000;
+    const LOG_EVERY_N_ATTEMPTS = 10;
+
     for (let attempt = 1; attempt <= MAX_RETRIES; attempt++) {
+      // Abort early if the analyzer process has already exited
+      if (this.analyzerRpcServer && this.analyzerRpcServer.exitCode !== null) {
+        throw new Error(
```
djzager: The `exit`/`close` handlers set `this.analyzerRpcServer = null`, so if the process exits during a retry delay, this guard becomes `null && ...` and the early-abort is skipped. The loop then retries for minutes against a dead process. Adding `!this.analyzerRpcServer` as an additional abort condition would catch this:
```typescript
if (!this.analyzerRpcServer || this.analyzerRpcServer.exitCode !== null) {
```

djzager (changes requested):
Nice approach — tailing the log file directly is the right call since errors go to analyzer.log, not stderr. Three things to address.

vscode/core/src/client/analyzerLogTailer.ts
abrugaro: Solved in 37ffde695dbbd790e7b82e11bda5c8d0ca251929

vscode/core/src/client/analyzerClient.ts
abrugaro: Solved in 37ffde695dbbd790e7b82e11bda5c8d0ca251929

vscode/core/src/client/analyzerClient.ts
abrugaro: Solved in 37ffde695dbbd790e7b82e11bda5c8d0ca251929

djzager (approved):
All three comments addressed: lastSize starts at 0, stderr filters progress JSON to debug, early-abort handles null server ref. Thanks.

PR #1464, opened 2026-06-24
:seedling: Move prompt templates into a governed, versioned package

fabianvf:
## What

Moves the LLM prompt strings that were inlined across the agentic nodes and `vscode/core` into a dedicated `@editor-extensions/prompts` workspace. Prompts are now Handlebars template files under `prompts/templates/`, rendered through a small `renderPrompt(id, context)` registry, and tracked by a checksummed `manifest.yaml`.

## Why

The built-in prompts are effectively the system's instruction set, but until now they were buried in compiled code with no review or validation story of their own. This is the engineering side of ISO 42001 A.5.2 (governance/lifecycle of AI systems): the prompts get version control, peer review, and CI validation independent of the surrounding code. Loops, conditionals, and per-language branching now live in the templates rather than in TypeScript.

## Behavior

No behavior change. The templates are reproduced from the previous strings **byte-for-byte** — there's a parity suite that renders every template and asserts exact equality against a verbatim copy of the old code, across each language / loop / conditional branch (including the existing whitespace quirks). The byte-exact output is also what ends up inlined in the built `extension.js`.

## What's here

- `prompts/` workspace — templates + partials and the render registry. Templates are embedded at build time, so `agentic` (tsup) and `vscode/core` (webpack) consume them without per-bundler config.
- `scripts/prompts-version.js` + `.github/workflows/prompt-validation.yml` — validate template syntax, placeholders, and manifest checksums on any `prompts/**` change, plus parity/drift and a deterministic (mock-model) semantic-regression check.
- `PROMPT_GOVERNANCE.md`, a CODEOWNERS rule for `/prompts/`, and an audit-evidence doc.

## Out of scope

The test-only `kai-evaluator` prompt, and the Kai solution-server prompts (owned in `konveyor/kai`).

## Needs a repo admin (can't be done in code)

- Create the `@konveyor/kai-prompt-reviewers` team.
- Branch protection on `main`/`release-*` requiring the `Validate prompt templates` check + Code-Owner review on `prompts/**` (documented in `PROMPT_GOVERNANCE.md`).

## Summary by CodeRabbit

* **New Features**
  * Added a governed `@editor-extensions/prompts` package with reusable Handlebars prompt templates (agentic analysis, diagnostics, quick actions, model health checks).
  * Added a CI workflow to validate prompt templates, manifest checksums, and deterministic rendered output on PRs.
* **Refactor**
  * Updated prompt construction to use shared template rendering for consistent system/human messaging.
* **Documentation**
  * Added `PROMPT_GOVERNANCE` and ISO evidence materials describing the prompt lifecycle and required review gates.
* **Tests**
  * Added byte-exact parity, snapshot drift, and semantic regression tests for prompt changes.


scripts/prompts-version.js
```diff
@@ -0,0 +1,177 @@
...
+      }
+    }
+
+    // Declared variables must actually be referenced inside a Handlebars tag
+    // (`{{ ... }}`) — a substring match would pass on a plain-text mention.
+    const tags = src.match(/{{[^}]*}}/g) ?? [];
+    for (const variable of entry.variables ?? []) {
+      const token = new RegExp(`\\b${variable.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")}\\b`);
+      if (!tags.some((tag) => token.test(tag))) {
+        problems.push(
```
djzager, line 209: The variable check is one-directional: it verifies manifest-declared variables appear in tags, but doesn't catch variables *used in templates* that aren't declared in the manifest. A typo like `{{migrationHnt}}` would silently render as empty and pass CI. Worth adding a reverse check that extracts identifiers from `{{...}}` tags and flags any that aren't declared, a registered helper, or a block-context keyword.

.github/CODEOWNERS
```diff
@@ -1,3 +1,9 @@
 * @konveyor/kai-ide-reviewers
 /tests @konveyor/kai-qe-reviewers @konveyor/kai-ide-reviewers
 
+# Governed prompt templates (ISO 42001 A.5.2). Any change to a prompt asset
+# requires sign-off from a designated Prompt Engineer / Product Architect.
+# See PROMPT_GOVERNANCE.md.
+/prompts/ @konveyor/kai-prompt-reviewers @konveyor/kai-ide-reviewers
+/scripts/prompts-version.js @konveyor/kai-prompt-reviewers @konveyor/kai-ide-reviewers
+
```
djzager, line 11: Consider also covering `PROMPT_GOVERNANCE.md` and `.github/workflows/prompt-validation.yml` here. Without those, someone could weaken the governance rules or disable the CI gate without prompt reviewer sign-off.

prompts/src/registry.ts
```diff
@@ -0,0 +1,43 @@
+// Import the CJS compiler build directly rather than "handlebars": the package's
+// Node entry registers `require.extensions` hooks that webpack (used to bundle
+// vscode/core) warns about and that we never use. This entry is the full
+// compiler (not runtime-only) in CJS, so it works under both esbuild and ts-jest.
+import Handlebars from "handlebars/dist/cjs/handlebars.js";
```
djzager, line 5: The rationale for this internal import makes sense, but since `handlebars/dist/cjs/handlebars.js` isn't part of the package's public API, consider pinning the exact `handlebars` version in `prompts/package.json` (instead of `^4.7.8`) to avoid a future release reorganizing its dist layout and breaking this silently.

prompts/src/registry.ts
```diff
@@ -0,0 +1,43 @@
...
+});
+
+for (const [name, source] of Object.entries(partialSources)) {
+  hb.registerPartial(name, source);
+}
+
+const compiled = new Map<PromptId, Handlebars.TemplateDelegate>();
+
+/** Render a governed Handlebars prompt template to its final string. */
+export function renderPrompt(id: PromptId, context: Record<string, unknown> = {}): string {
```
djzager: nit: Since each `PromptId` has known required variables (they're in the manifest), there's an opportunity to add a type map so callers get compile-time checking on context keys. Fine as a follow-up, but worth considering — a typo like `programmingLangage` silently renders as empty with no error today.

scripts/prompts-version.js
```diff
@@ -0,0 +1,177 @@
...
+      console.error("  - " + p);
+    }
+    console.error(`\n${problems.length} problem(s) found.`);
+    process.exit(1);
+  }
+
+  console.log(
+    `Prompt template validation passed. ` +
+      `version=${manifest.version}, templates=${(manifest.templates ?? []).length}, ` +
+      `partials=${(manifest.partials ?? []).length}.`,
```
djzager, line 245: The `update` path does a raw `readFileSync` with no error handling. If a manifest entry references a path that doesn't exist on disk (e.g., after a rename), this crashes with an unhelpful ENOENT. A try/catch with a clear message about which manifest entry failed would improve the DX.

scripts/prompts-version.js
```diff
@@ -0,0 +1,177 @@
...
+// Usage:
+//   node scripts/prompts-version.js check     # CI gate (exits non-zero on any problem)
+//   node scripts/prompts-version.js update    # recompute + write checksums (dev convenience)
+
+import { createHash } from "node:crypto";
+import { readFileSync, writeFileSync, readdirSync, statSync } from "node:fs";
+import { join, relative, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+import yaml from "js-yaml";
+import Handlebars from "handlebars";
```
djzager, line 21: `handlebars` is imported here but is only declared as a dependency in `prompts/package.json`, not the root. This works via npm workspace hoisting today, but is fragile under stricter package managers or different hoisting behavior. Consider adding `handlebars` to root `devDependencies`, or running this script from within the `prompts` workspace.

djzager (changes requested):
Overall this is well-structured — the byte-exact parity testing with the oracle is a strong design choice, and the architecture is sound. A few substantive issues below.

djzager (approved):
All comments addressed — reverse variable check, CODEOWNERS coverage, pinned handlebars, type-safe context map, update() error handling, and version lockstep enforcement. Nice work.

PR #1465, opened 2026-06-29
:seedling: Auto-create cherry-pick label in cut-release workflow

djzager:
Two fixes to the cut-release workflow's post-release housekeeping:

1. **Create cherry-pick label automatically** — adds a `cherry-pick/release-X.Y` label so cherry-pick automation works immediately without manual label creation.

2. **Use a PR instead of direct push to main** — the housekeeping step was pushing directly to `main`, which fails due to branch protection rules (`GH013: Changes must be made through a pull request`). Now it creates a `post-release/vX.Y.0` branch and opens a PR with the changelog assembly and version bump.

### DIFF
--- a/.github/workflows/cut-release.yml
+++ b/.github/workflows/cut-release.yml
@@ -169,6 +169,15 @@ jobs:
 
           git push origin ${{ needs.validate.outputs.branch }}
 
+      - name: Create cherry-pick label
+        run: |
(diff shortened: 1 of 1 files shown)


savitharaghunathan (approved):
visiack

fabianvf (approved):

Issue #1273, opened 2026-02-25
[Solution Server] Incident status is not updated when an action is taken between token refreshes


abrugaro:
When a user requests a new fix, it is created on the solution server with a pending status. If the solution server token expires and is refreshed before the user takes any action (accept/reject), the solution may remain in the pending status indefinitely, even after the user performs an action.

### Steps to reproduce

1. Connect to the solution server, analyze an application and request a fix (the incident is created in the solution server with the pending status)
2. Wait until the token needs to be refreshed (the "successfully connected to the solution server" notification is displayed again)
3. Take any action on the proposed solution (Accept/Reject)

 ### Env
- Latest dev build
- Coolstore app
- Solution server deployed from main

### More info

```json
{
  "counted_solutions": 4,
  "accepted_solutions": 2,
  "rejected_solutions": 0,
  "modified_solutions": 0,
  "pending_solutions": 2, <--------
  "unknown_solutions": 0
}
```

<img width="747" height="523" alt="Image" src="https://github.com/user-attachments/assets/e9984b29-f2ff-4193-91e0-ccf2f00217a3" />

RanWurmbrand:
Hi! It looks like this issue is a duplicate of #1150 
i did some investigation over there and it seems to be caused by a race condition in SolutionServerClient during token updates. hope that help :)

PR #1468, opened 2026-07-02
:bug: Update solution server status on stale MCP connection loss

ibolton336:
## Summary

Fixes #1433 — the Solution Server status chip stayed green after the MCP SSE connection went stale.

There were three compounding problems:

1. **No state propagation**: `SolutionServerClient` detected connection failures, set `isConnected = false`, and closed the stale client internally — but never notified `HubConnectionManager`, so `solutionServerConnected` was never broadcast to the webview.
2. **Health poll masked failures**: the 10s health poll calls `getServerCapabilities(true)`, but on connection errors that method swallowed the error and returned empty capabilities. The poll's success path then ran and set the chip back to green *right after* the client had detected the failure.
3. **Polling stopped permanently** after 5 consecutive failures with no way to resume (the manual restart command doesn't restart the poll loop), so the state could never recover automatically.

## Changes

**`agentic/src/clients/solutionServerClient.ts`**
- Add `setConnectionStateListener()` — fires on every `isConnected` transition (connect, disconnect, and stale-connection detection in `getServerCapabilities`/`getBestHint`)
- `getServerCapabilities()` now throws `SolutionServerClientError` on connection failure instead of returning empty capabilities, so the health poll registers the disconnect
- Classify undici `UND_ERR_*` codes (e.g. the `UND_ERR_CONNECT_TIMEOUT` from the issue's logs) as connection errors

**`vscode/core/src/hub/HubConnectionManager.ts`**
- Wire every created `SolutionServerClient` to a new `setSolutionServerConnectionCallback()` so state changes propagate out
- New `reconnectSolutionServer()`: lightweight reconnect that reuses the current bearer token without re-authenticating or disturbing other Hub features; triggers workflow disposal on success so the next workflow run picks up the new client instance

**`vscode/core/src/extension.ts`**
- Broadcast connection state changes to the webview (`solutionServerConnected`) — the chip now goes grey the moment a stale connection is detected, including mid-workflow
- Health poll auto-reconnects on failure and backs off to a slow 5-minute heartbeat instead of stopping permanently

## Behavior after this change

- Idle timeout (~18–30 min) → next request or poll detects the failure → chip turns grey immediately
- The poll then attempts reconnection automatically (10s → 30s → 60s → 5m backoff) and the chip returns to green once the connection is re-established — no manual restart needed

## Also: token refresh no longer replaces Hub client instances

While wiring the above, we found the other trigger of the same stale-client family: `refreshTokens()` called `reconnectClients()`, which built **new** `SolutionServerClient`/`ProfileSyncClient` instances on every token refresh — without disconnecting the old ones and without firing `onWorkflowDisposal`. This is a regression of #1273's fix (#1283, "don't nuke clientid for token refresh"), reintroduced by the OIDC rework in #1419 and shipping since v0.5.25. It only bites configs where the refresh timer actually arms (credentials auth, or OIDC when the PAT exchange fails — the happy OIDC path uses a long-lived PAT and never refreshes), but there the consequences compound: a mid-run workflow keeps an orphaned client whose token dies ~30s post-refresh (refresh fires `TOKEN_EXPIRY_BUFFER_MS` early), batch-review accept/reject goes out with an empty `client_id` (breaking success-rate correlation), the orphaned client's still-attached state listener flaps the status chip, and the hub-proxy model provider — whose token is baked into its `ChatOpenAI` instances — 401s every subsequent solution run with no auto-recovery.

**`vscode/core/src/hub/HubConnectionManager.ts`**
- `reconnectClients()` now updates the existing instances in place via `updateBearerToken()` (restoring the #1283 pattern), preserving workflow references and `clientId` session state; a new instance is created only when a client is missing, and on update failure the instance is kept so the health poll above can recover it
- `refreshTokens()` now fires `onWorkflowDisposal(tokenRefreshOnly)` after a successful refresh, so the extension rebuilds the model provider with the fresh token — `extension.ts`'s existing `tokenRefreshOnly` handling (skip workflow disposal, rebuild provider) finally gets exercised as its comment always claimed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Solution Server status handling so connection changes are reflected immediately in the webview and the app can recover automatically.
  * Fixed stale connection behavior by continuing health polling with backoff and attempting reconnection after capability-load failures.
  * Reduced issues during token refresh by preserving existing client state when possible, preventing session churn and avoiding post-refresh 401 failures.
  * Prevented token refresh loops when no expiry is known, and improved reliability of capability/hint retrieval error handling.

### DIFF
--- a/agentic/src/clients/solutionServerClient.ts
+++ b/agentic/src/clients/solutionServerClient.ts
@@ -110,6 +110,9 @@ export class SolutionServerClientError extends Error {
   }
 }
 
+/** Listener invoked whenever the client's connection state transitions. */
+export type SolutionServerConnectionListener = (connected: boolean) => void;
+
 /**
  * Client for interacting with the MCP solution server.
  *
@@ -124,6 +127,7 @@ export class SolutionServerClient extends KaiWorkflowEventEmitter {
   private currentClientId: string = "";
   private logger: Logger;
   private cachedCapabilities: SolutionServerCapabilities | null = null;
+  private connectionStateListener: SolutionServerConnectionListener | null = null;
   public isConnected: boolean = false;
 
   constructor(
@@ -140,6 +144,24 @@ export class SolutionServerClient extends KaiWorkflowEventEmitter {
     this.logger = logger.child({
       component: "SolutionServerClient",
     });
+  }
+
+  /**
+   * Register a listener that fires whenever the connection state changes.
+   * Used by HubConnectionManager to propagate disconnects (e.g. stale MCP
+   * connections detected mid-request) to the webview status UI.
+   */
+  public setConnectionStateListener(listener: SolutionServerConnectionListener | null): void {
+    this.connectionStateListener = listener;
+  }
+
+  /** Set isConnected and notify the listener when the value transitions. */
+  private setConnected(connected: boolean): void {
+    const changed = this.isConnected !== connected;
+    this.isConnected = connected;
+    if (changed) {
+      this.connectionStateListener?.(connected);
+    }
   }
 
   public async connect(): Promise<void> {
@@ -211,7 +233,7 @@ export class SolutionServerClient extends KaiWorkflowEventEmitter {
         new StreamableHTTPClientTransport(serverUrlObj, transportOptions),
       );
       this.logger.info("Connected to MCP solution server");
-      this.isConnected = true;
+      this.setConnected(true);
       return true;
     } catch (error) {
       this.logger.warn(`Failed to connect with original URL (${this.serverUrl})`, error);
@@ -239,7 +261,7 @@ export class SolutionServerClient extends KaiWorkflowEventEmitter {
         this.logger.info(
           `Connected to MCP solution server using alternative URL: ${alternativeUrl}`,
         );
-        this.isConnected = true;
+        this.setConnected(true);
         return true;
       } catch (error) {
         this.logger.error(`Failed to connect with alternative URL (${alternativeUrl})`, error);
@@ -269,7 +291,7 @@ export class SolutionServerClient extends KaiWorkflowEventEmitter {
         this.mcpClient = null;
       }
 
-      this.isConnected = false;
+      this.setConnected(false);
       this.logger.info("Disconnected from MCP solution server");
     } catch (error) {
       this.logger.error("Error during disconnect", error);
@@ -361,7 +383,8 @@ export class SolutionServerClient extends KaiWorkflowEventEmitter {
         errorMessage.toLowerCase().includes("etimedout") ||
         errorCode === "ECONNRESET" ||
         errorCode === "ECONNREFUSED" ||
-        errorCode === "ETIMEDOUT";
+        errorCode === "ETIMEDOUT" ||
+        (typeof errorCode === "string" && errorCode.startsWith("UND_ERR"));
 
       if (isConnectionError) {
         // Connection error - log detailed diagnostic information
@@ -375,17 +398,17 @@ export class SolutionServerClient extends KaiWorkflowEventEmitter {
           fullError: JSON.stringify(error, Object.getOwnPropertyNames(error)),
         });
 
-        // Mark as disconnected
-        this.isConnected = false;
+        // Mark as disconnected and notify the connection state listener
+        this.setConnected(false);
 
         // Close the stale MCP client to prevent resource leaks
         await this.closeStaleClient();
 
-        // Return empty capabilities instead of throwing
-        return {
-          tools: [],
-          resources: [],
-        };
+        // Propagate the failure so callers (e.g. the health poll) register
+        // the disconnect instead of mistaking empty capabilities for success
+        throw new SolutionServerClientError(
+          `Solution server connection failure: ${errorCode ?? errorMessage}`,
+        );
       }
 
       this.logger.error("Failed to get server capabilities", error);
@@ -793,7 +816,8 @@ export class SolutionServerClient extends KaiWorkflowEventEmitter {
         errorMessage.toLowerCase().includes("network") ||
         errorCode === "ECONNRESET" ||
         errorCode === "ECONNREFUSED" ||
-        errorCode === "ETIMEDOUT";
+        errorCode === "ETIMEDOUT" ||
+        (typeof errorCode === "string" && errorCode.startsWith("UND_ERR"));
 
       if (isNotFoundError) {
         // Treat "not found" as a normal case - the violation simply has no hint in the database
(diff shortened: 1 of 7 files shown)


djzager (commented):
Good approach to a tricky problem — the three fixes (state propagation, health poll recovery, token refresh client preservation) address the root causes well. A few concerns about concurrency and interaction effects below.

vscode/core/src/hub/HubConnectionManager.ts
```diff
@@ -504,57 +531,142 @@ export class HubConnectionManager {
...
+    if (!this.config.enabled || !this.config.features.solutionServer.enabled) {
+      return false;
+    }
+    if (this.config.auth.enabled && !this.bearerToken) {
+      // No usable token — reconnecting requires re-authentication (sign in)
+      return false;
+    }
+
+    const staleClient = this.solutionServerClient;
+    this.solutionServerClient = null;
```
djzager: `reconnectSolutionServer()` nulls out `this.solutionServerClient` before the stale client is disconnected and before the new client is connected — exposing a window where concurrent callers (health poll, workflows) see `null`. Unlike `connect()` which has `connectPromise` as a guard, there's no reconnect mutex here. Two concurrent health poll failures could both call `reconnectSolutionServer()` and race. Consider adding a reconnect guard, or only replacing the reference atomically after the new client is confirmed connected.

agentic/src/clients/solutionServerClient.ts
```diff
@@ -142,6 +146,24 @@ export class SolutionServerClient extends KaiWorkflowEventEmitter {
...
+   * Used by HubConnectionManager to propagate disconnects (e.g. stale MCP
+   * connections detected mid-request) to the webview status UI.
+   */
+  public setConnectionStateListener(listener: SolutionServerConnectionListener | null): void {
+    this.connectionStateListener = listener;
+  }
+
+  /** Set isConnected and notify the listener when the value transitions. */
+  private setConnected(connected: boolean): void {
+    const changed = this.isConnected !== connected;
```
djzager, line 161: `updateBearerToken` disconnects then reconnects, firing the connection state listener twice — the status chip will briefly flash grey on every token refresh. Consider suppressing the listener during the disconnect/reconnect cycle (e.g., `this.connectionStateListener = null` around the calls, then restore and fire only if the final state differs from the initial state).

vscode/core/src/hub/HubConnectionManager.ts
```diff
@@ -504,57 +531,142 @@ export class HubConnectionManager {
...
+          this.logger.info("Updating solution server client with refreshed token");
+          await this.solutionServerClient.updateBearerToken(this.bearerToken);
+          if (!this.solutionServerClient.isConnected) {
+            await this.solutionServerClient.connect();
+          }
+        } catch (error) {
+          // Keep the instance — the health poll (reconnectSolutionServer) can
+          // recover the connection without stranding workflow references
+          this.logger.error("Failed to update solution server client token", error);
+        }
```
djzager, line 593: If `updateBearerToken` fails mid-way (disconnect succeeds, reconnect fails), the client is left disconnected but `solutionClientReplaced` stays `false`, so workflows aren't disposed. However, the health poll's `reconnectSolutionServer()` creates a *new* instance anyway, orphaning the workflow's reference. Consider setting `solutionClientReplaced = true` in the catch block so workflows get disposed and pick up the eventually-reconnected client.

vscode/core/src/hub/HubConnectionManager.ts
ibolton336: Fixed in a0c17784. Went a bit further than a mutex after tracing the interleavings:

- Concurrent calls now coalesce behind a `reconnectPromise` mutex (same pattern as `connectPromise`).
- The replacement client is built and connected **before** the swap, so readers never observe a null client mid-reconnect; the stale (already disconnected) instance stays in place until then.
- The swap re-checks every invalidating condition (`connectPromise`, logout teardown, in-flight token refresh, a `connectionEpoch` bumped by any manager-level disconnect, cleared bearer token) and discards the freshly-built client if anything superseded it — this closes the nastier races where an in-flight reconnect could install a client built from a pre-`updateConfig` URL or resurrect a connection after `oidcLogout` revoked the PAT.
- The workflow-disposal callback is wrapped so a throw can't reject the reconnect and permanently kill the health poll's rescheduling.

agentic/src/clients/solutionServerClient.ts
ibolton336: Fixed in a0c17784. Implemented with a reentrancy-safe suppression **counter** rather than null-and-restore of the listener field: `setConnected` only notifies at depth 0, and `updateBearerToken`'s `finally` fires through the *current* listener field only when the final state differs from the initial one.

The counter approach matters for two interleavings the restore version gets wrong: (1) concurrent `updateBearerToken` calls could restore a stale capture and permanently wipe the listener; (2) if the health poll replaces the client while its `updateBearerToken` is suspended at an await, a restore would re-attach the listener to the orphaned instance and broadcast a stale `disconnected` over the new client's `connected`.

Verified against a local mock hub: token refresh with the server up fires zero events (no grey flash); with the server down it fires exactly one `false`.

vscode/core/src/hub/HubConnectionManager.ts
ibolton336: Fixed in a0c17784 — the catch now sets `solutionClientReplaced = true` exactly as suggested, so a mid-refresh failure triggers full workflow disposal instead of leaving a workflow holding the client the health poll is about to replace. Also made `refreshTokens` drain any in-flight poll reconnect before touching the clients (and vice versa: the reconnect's swap point re-checks `isRefreshingTokens`), so the two recovery paths can't tear down each other's instances.

vscode/core/src/hub/HubConnectionManager.ts
```diff
@@ -655,6 +849,21 @@ export class HubConnectionManager {
...
   public async oidcLogout(): Promise<void> {
+    // Block in-flight/new health-poll reconnects for the whole teardown — a
+    // reconnect completing mid-logout would resurrect a connection with the
+    // PAT the user is revoking
+    this.teardownInProgress = true;
+    try {
+      if (this.reconnectPromise) {
+        await this.reconnectPromise.catch(() => {});
+      }
+      await this.doOidcLogout();
```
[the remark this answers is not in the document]
ibolton336: Fixed in ea27f110 — logout now drains an in-flight `connectPromise` (as suggested) before tearing down. Also applied the two outside-diff findings from the same review: partial-client cleanup in `_connect`/`reconnectClients` failure paths, and the guarded workflow-disposal callback in the token-refresh path.

fabianvf (approved):
LGTM

djzager (approved):
All three concerns addressed — reconnect mutex with atomic swap, listener suppression counter for token refresh, and catch block now signals workflow disposal. Nice turnaround.

PR #1470, opened 2026-07-08
:sparkles: Add native Anthropic API provider (ChatAnthropic)

fabianvf:
## Summary

Adds first-class support for calling the **Anthropic Messages API directly** via LangChain's `ChatAnthropic`, alongside the existing providers.

Previously, `ChatAnthropic` was not a supported provider — `getModelProviderFromConfig` rejects any provider string not in the `ModelCreators` registry with `"Unsupported model provider"`. Claude models were only reachable **indirectly**:
- via **AWS Bedrock** (`ChatBedrockConverse`, through Bedrock's Converse API), or
- via **`ChatOpenAI`** pointed at an OpenAI-compatible proxy `baseURL`.

This PR wires up the native path so users can drop Anthropic settings straight into `provider-settings.yaml`.

## Changes

- **`modelCreator.ts`** — register a `ChatAnthropic` creator in the `ModelCreators` registry, routed through the same unified TLS / custom-fetch setup as the other providers. The custom fetch (for CA bundles / insecure mode) is injected via the Anthropic SDK's `clientOptions.fetch`.
- **`types.ts` (client + modelProvider)** — add `"ChatAnthropic"` to the `SupportedModelProviders` union.
- **`sample-provider-settings.yaml`** — document an `Anthropic` block.
- **`package.json`** — add `@langchain/anthropic`.

## Sample config

```yaml
  Anthropic:
    environment:
      ANTHROPIC_API_KEY: "" # Required
    provider: ChatAnthropic
    args:
      model: claude-sonnet-5 # Required
      # anthropicApiUrl: <string> # override the API base URL
      # clientOptions: <object>   # passed through to the Anthropic SDK client
```

## Dependency note

`@langchain/anthropic` is pinned to `~1.1.3` to match the repo's pinned `@langchain/core` (`1.1.x`, enforced repo-wide via the `overrides` field). Newer `@langchain/anthropic` lines (1.3+) require `@langchain/core` ≥ 1.1.45 / 1.2.x and fail to bundle against the current core (missing `ContextOverflowError` / `language_models/structured_output` exports). Bumping core would ripple across all providers, so pinning Anthropic to the compatible line is the surgical choice; a follow-up can bump both together. The lockfile change adds only `@langchain/anthropic@1.1.3` and `@anthropic-ai/sdk@0.71.2` — no other provider's resolved version shifts.

## Testing

- `tsc --noEmit` passes.
- `webpack --mode production` (core) builds cleanly with the new dependency bundled.
- eslint/prettier clean.

Runtime model calls require a live `ANTHROPIC_API_KEY` and were not exercised in CI; the wiring mirrors the existing providers exactly.

### DIFF
--- a/changes/unreleased/1470-native-anthropic-provider.yaml
+++ b/changes/unreleased/1470-native-anthropic-provider.yaml
@@ -1,0 +1,5 @@
+kind: feature
+description: >
+  Added native Anthropic API support via a ChatAnthropic model provider, so Claude
+  models can be configured directly instead of only through AWS Bedrock or an
+  OpenAI-compatible proxy.
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,6 +22,7 @@
         "vscode/konveyor"
       ],
       "dependencies": {
+        "@langchain/anthropic": "~1.1.3",
         "@langchain/aws": "^1.2.2",
         "@langchain/core": "^1.1.7",
         "@langchain/deepseek": "^1.0.11",
@@ -112,6 +113,26 @@
     "extra-types": {
       "name": "@types/extra-types",
       "version": "0.6.0"
+    },
+    "node_modules/@anthropic-ai/sdk": {
+      "version": "0.71.2",
+      "resolved": "https://registry.npmjs.org/@anthropic-ai/sdk/-/sdk-0.71.2.tgz",
+      "integrity": "sha512-TGNDEUuEstk/DKu0/TflXAEt+p+p/WhTlFzEnoosvbaDU2LTjm42igSdlL0VijrKpWejtOKxX0b8A7uc+XiSAQ==",
+      "license": "MIT",
+      "dependencies": {
+        "json-schema-to-ts": "^3.1.1"
+      },
+      "bin": {
+        "anthropic-ai-sdk": "bin/cli"
+      },
+      "peerDependencies": {
+        "zod": "^3.25.0 || ^4.0.0"
+      },
+      "peerDependenciesMeta": {
+        "zod": {
+          "optional": true
+        }
+      }
     },
     "node_modules/@asamuzakjp/css-color": {
       "version": "3.2.0",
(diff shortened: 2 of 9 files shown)


ibolton336 (commented):
## Code Review

**Overall: Clean, well-structured PR that follows established project conventions precisely. No correctness bugs found.**

### Strengths

- **Consistent with existing patterns.** `ChatAnthropicCreator` follows the exact same `ModelCreator` interface, TLS setup, and validation logic as every other provider.
- **Correct TLS/fetch injection.** The Anthropic SDK uses `clientOptions.fetch` (not `configuration.fetch` like OpenAI) — the PR correctly targets that path.
- **Good test coverage.** Three TLS tests (self-signed cert success, no-cert failure, insecure bypass) mirror the OpenAI and Google GenAI test triplets. Mock server endpoint at `/v1/messages` returns a valid Anthropic Messages API response shape.
- **Thoughtful `defaultArgs`.** No `temperature` default since current Claude models reject sampling params with a 400 — the comment explains why.
- **Dependency pinning rationale.** `~1.1.3` is well-justified to avoid a `@langchain/core` version cascade.

### Minor Suggestions

1. **`apiKey` can be silently overridden by `args` spread** — In `create()`, `...args` comes after `apiKey: env.ANTHROPIC_API_KEY`, so a user who accidentally puts `apiKey` in their YAML args would shadow the env variable. This matches the existing providers (OpenAI, Azure, DeepSeek all do the same), so it's consistent — just worth being aware of.

2. **Sample config model ID** — `sample-provider-settings.yaml` uses `model: claude-sonnet-5`. Worth verifying this matches the current canonical model ID (e.g., `claude-sonnet-5-20250514` vs short alias) so users copy-pasting the sample don't hit errors.

3. **No streaming mock** — The mock endpoint only handles non-streaming, and the test config sets `streaming: false`. Since the tests are TLS-focused this is fine, but a streaming mock could be a useful follow-up.

4. **Lockfile churn** — The large lockfile diff is just npm reclassifying `peer: true` annotations. Actual new packages are only `@langchain/anthropic@1.1.3`, `@anthropic-ai/sdk@0.71.2`, `json-schema-to-ts@3.1.1`, and `ts-algebra@2.0.0`. No version changes to existing deps — looks safe.

ibolton336 (approved):

PR #1477, opened 2026-07-28
:bug: Validate custom rules file types in profile file dialog

ibolton336:
Fixes #580

## Summary

The "Select Custom Rules" dialog (Manage Profile → Configure Custom Rules) accepted any file type — `.java`, `.png`, etc. were silently added to the profile as custom rules.

## Changes

- Filter the file dialog to YAML rule files (`.yaml`/`.yml`), matching the webview upload path which already enforces YAML-only
- Validate the selection after the dialog closes, since dialog filters don't apply to folders and aren't enforced on all platforms: non-YAML files are skipped with a warning listing the skipped names
- Folders are still accepted as rule directories
- Adds a changelog fragment

## Verification

- `tsc --noEmit` clean on `vscode/core`
- `eslint` clean on the changed file
- `npm run changelog:validate` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Custom rule selection now accepts YAML rule files and directories while ignoring unsupported file types.
  - Invalid selections are skipped with a warning, helping prevent unintended files from being added.
  - The configuration flow now stops when no valid rule inputs remain.

PR #1480, opened 2026-08-04
:seedling: unbreak vscode-test on VS Code 1.131.0 and the changelog check

fabianvf:
## Summary

`Test (macos)` has been failing on every PR since VS Code 1.131.0 shipped, and `Verify changelog fragment` fails on every fork PR. Both are broken on main, not by any particular change. Confirmed by running `npx vscode-test` on a clean `ed42f5a` checkout.

- 1.131.0 dropped `Contents/MacOS/Electron` for `Contents/MacOS/Code`, and `@vscode/test-electron` 2.x still spawns the old name. `.vscode-test.mjs` pins no version so it picks up stable at run time, which is why this started on its own. #1477 passed on 1.130.0. macOS failing then cancels linux and windows via fail-fast, so all three go red.
- Bump `@vscode/test-electron` to 3.x. That gets it launching and exposes a second macOS failure underneath: VS Code opens a unix socket in its user data dir, and the default `.vscode-test/user-data` path is over the 104 byte `sun_path` limit. CI's macOS path is exactly 104, so the bump on its own would have swapped one failure for another. Point `--user-data-dir` at a short temp path.
- `@vscode/test-cli` stops at 0.0.14 on purpose. 0.0.15 pulls `yargs@18`, which wants node `^22.12.0`, and `.nvmrc` is v22.9.0, so install dies with EBADENGINE before anything runs. Bumping `.nvmrc` would also work but that is a bigger call than this PR should make.
- The changelog job checks out the PR head, which `actions/checkout` now refuses under `pull_request_target` since that runs fork code with a privileged token. It only needs the changed filenames, so ask the API for them instead. Avoids opting into `allow-unsafe-pr-checkout`.

`test:integration` reports 0 passing before and after, because the webpack build never emits `out/test/`. That step exists to launch VS Code and exit clean, and it does again. The 150 tests CI reports come from `test:unit-tests`.

## Test plan

- [x] `npx vscode-test` on main reproduces the macOS failure, and passes with these changes
- [x] `Test (macos)`, `Test (linux)` and `Test (windows)` green on this PR
- [ ] `Verify changelog fragment` green on the next bugfix or feature PR. It is skipped here since this is `:seedling:`


djzager (approved):
ACK

PR #1487, opened 2026-08-19
:test_tube: De-flake tier1 'Auto Accept on Save' tab close

djzager:
## Problem

The `tier1` job has been failing frequently on both `main` and `release-0.6`, blocking patch releases. The culprit is a single flaky test:

`e2e/tests/base/plugin-settings.test.ts › Plugin Settings - Analyze on Save › Enable "Auto Accept on Save" setting`

It fails in `TabManager.closeTabByName('CatalogService.java')` at the close-button step. The failure mode **varies run-to-run**, which is the tell for a timing/render race rather than a code regression:

- **Element not found** — Patch Release run [32236624704](https://github.com/konveyor/editor-extensions/actions/runs/32236624704/job/96073936896) (main): `expect(closeBtn).toBeVisible()` → `<element(s) not found>`.
- **Found but not clickable** — CI run 31706377411 (main) and 32149593086 (release-0.6): the button resolves to `<a class="action-label codicon codicon-close">` but `.click()` times out.

### Root cause

`ReviewInEditor` opens an inline vertical diff. On `File: Save`, the extension auto-accepts and mutates the buffer, and the test immediately calls `closeTabByName`. The tab's close button only renders reliably for the **active** tab; for an inactive tab the `.tab-actions` subtree can be briefly detached while VS Code re-renders during the dirty → clean transition. With `retries: 0`, a single such flake reds the whole (release-blocking) tier.

## Changes

- **`closeTabByName`**: focus the tab first (so its close button is reliably rendered), fall back to the `View: Close Editor` command if the DOM click races a re-render, and assert the tab is actually gone.
- **`playwright.config.ts`**: `retries: process.env.CI ? 2 : 0` so a single flaky interaction doesn't red the tier; local runs keep fast 0-retry feedback.

## Verification

Can't reliably run the full Electron E2E stack locally, so this needs to bake in CI — please watch the `tier1` job. `tsc`/`eslint` clean for the changed files (only pre-existing warnings remain).

The related product bug (auto-accept-on-save passing `fsPath` instead of the URI string) is fixed in a separate `:bug:` PR.

### DIFF
--- a/changes/unreleased/1488-auto-accept-on-save-uri.yaml
+++ b/changes/unreleased/1488-auto-accept-on-save-uri.yaml
@@ -1,0 +1,9 @@
+kind: bugfix
+description: >
+  Fix "Auto Accept on Save" saving files with un-accepted inline diffs. The
+  save handler looked up the diff by its URI string but then tried to accept
+  it by filesystem path, so the accept silently no-oped and the file was saved
+  with the pending diff still in the buffer. The correct URI is now used, so
+  saving accepts and clears the diff as intended.
+extensions:
+  - core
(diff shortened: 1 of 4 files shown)


fabianvf (approved):
LGTM

